PR #211, opened 2019-11-20
Delete README.rst

PatMyron:
https://github.com/aws-cloudformation/aws-cloudformation-rpdk-java-plugin/blob/master/README.md


rjlohan:
Thanks, we added this to publish to PyPi before we realised we could just set the markdown format. :)

PR #215, opened 2019-12-05
Adding Travis CI

jotompki:
*Issue #, if available:*

*Description of changes:* Adding TravisCI job that should have parity with current codebuild job. 
Planning on adding deployment stages and maybe separating out linting from unit tests, but this seems like a good start. There are four testing jobs that run concurrently, three for the versions of python we support and one for the java portion of the plugin. If these are all successful, an integration stage runs that checks that the two can integrate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,45 @@
+dist: bionic
+jdk: openjdk11
+cache:
+  - pip
+  - directories:
+    - $HOME/.m2
+env:
+  global:
+  - AWS_REGION="us-east-1"
+  - AWS_DEFAULT_REGION=$AWS_REGION
+before_install:
+  - >
+    pip install --upgrade pip 'attrs==19.2.0'
+    -r https://raw.githubusercontent.com/aws-cloudformation/cloudformation-cli/master/requirements.txt
+install:
+  - pip install .
+
+script:
+  - pre-commit run --all-files
+
+jobs:
+  include:
+    - language: python
+      python: "3.6"
+    - language: python
+      python: "3.7"
+    - language: python
+      python: "3.8"
+    - language: java
+      before_install: skip
+      install: skip
+      script: mvn verify
+    - stage: integ
+      language: python
+      python: "3.6"
+      install:
+        - mvn install
+        - pip install .
+      script:
+        - DIR=$(mktemp -d)
+        - cd "$DIR"
+        - ls -la
+        - printf "AWS::Foo::Bar\n\n" | cfn init -vv
+        - mvn verify
+        - ls -la
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,12 +47,6 @@ repos:
     files: "^python/"
 - repo: local
   hooks:
-  - id: git-secrets
-    name: git-secrets
-    description: Run git-secrets
-    entry: git-secrets --scan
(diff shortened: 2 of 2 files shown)


tobywf:
Err, what? Doco clearly says [to use AWS_DEFAULT_REGION](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html). Might be a different issue, in that we're not setting the region properly in a test, and we haven't noticed due to developer machines and CodeBuild having it.

jotompki:
Yeah this is being thrown when a [test attempts to create a client](https://travis-ci.com/aws-cloudformation/cloudformation-cli-java-plugin/jobs/263680805#L714). I guess on completely clean non ec2 environments this isn't there

tobywf:
fair enough, that's a bit naff. might be good to have both in there? the CLI and boto defo honour the other one

pom.xml
```diff
@@ -363,37 +363,6 @@
                     <failOnViolation>true</failOnViolation>
                 </configuration>
             </plugin>
-            <plugin>
```
tobywf: ?

.travis.yml
```diff
@@ -0,0 +1,48 @@
...
+      language: python
+    - python: "3.7"
+      language: python
+    - python: "3.8"
+      language: python
+    - language: java
+      before_install: skip
+      install: skip
+      script: mvn verify
+    - stage: integ
```
tobywf: doesn't this step need a python version specified? getting:

>  This job ran using the default Python version, which was updated to 3.6 on April 16th, 2019. You can explicitly stay on the previous version by specifying python: 2.7 in your .travis.yml.

might be good to pin it

.travis.yml
```diff
@@ -0,0 +1,48 @@
...
+      script:
+        - DIR=$(mktemp -d)
+        - cd "$DIR"
+        - ls -la
+        - printf "AWS::Foo::Bar\n\n" | cfn init -vv
+        - mvn verify
+        - ls -la
+    - stage: deploy-maven
+      if: branch = master
+      script: skip
```
tobywf: can we remove this for now? it still takes 1+ minutes to spin up the worker, even with the skip condition

pom.xml
jotompki: sorry iterating on this just testing some things out with travis might have snuck in

.travis.yml
jotompki: good idea will do

.travis.yml
jotompki: yep was just trying some stuff out wasn't final

.travis.yml
```diff
@@ -0,0 +1,45 @@
+dist: bionic
+jdk: openjdk11
```
PatMyron: do we want to support multiple Java versions?

tobywf: JDK8 is ancient, and is unmaintained by Oracle. OpenJDK8 is still maintained. but we can set a compatible build even in newer JDKs:

https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/61f736d2650dd03a954cdd72ac7b2028b93a7ce9/pom.xml#L36-L37

so i think it's fine. i'd rather stick with the Bionic Ubuntu LTS (18.04) and OpenJDK 11 than some older build env.

PR #240, opened 2020-04-03
added aws-guided codegen model 

ammokhov:
*Issue #, if available:*

*Description of changes:*

along with @rjlohan refactored codegen model to allow `aws guided` option. This option generates resource handlers (each handler has a special stub) and supporting files allowing developer to use functional paradigm.
refactored files:

- `BaseHandlerStd` - abstract class that has a `handleRequst` wrapper
- `CreateHandler` - has a sample create handler context pattern
- `DeleteHandler` - has a sample delete handler context pattern
- `ReadHandler` - has a sample read handler context pattern
- `UpdateHandler` - has a sample update handler context pattern
- `ClientBuilder` - has a sample sdkClient builder
- `Translator` - has sample api requests

Having progress chain pattern pre-generated for developers gives less function state control, less worry of exception handling, and more standardized approach for any future custom resource implementations. 

--------------------------
```bash
% cfn init
% Initializing new project
% What's the name of your resource type?
% (Organization::Service::Resource)
>> 
% One language plugin found, defaulting to java
% Enter a package name (empty for default 'com.aa.aa.aa'):
>>
% Choose codegen model - 1 (default) or 2 (guided-aws):
```
--------------------------

`mvn verify` builds successfully:
```
[INFO] All coverage checks have been met.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```


aygold92:
please update the name of this PR

aygold92:
can you please post the results of a sample code generation?

python/rpdk/java/codegen.py
```diff
@@ -10,12 +11,30 @@
...
 from .resolver import translate_type
-from .utils import safe_reserved, validate_namespace
+from .utils import safe_reserved, validate_codegen_model, validate_namespace
 
 LOG = logging.getLogger(__name__)
 
 OPERATIONS = ("Create", "Read", "Update", "Delete", "List")
 EXECUTABLE = "cfn"
+AWSCODEGEN = namedtuple("AWSCODEGEN", "default guided default_code guided_code")
+CODEGEN = AWSCODEGEN("default", "guided_aws", "1", "2")
```
surega21, line 21: what is default here ?

python/rpdk/java/templates/init/guided_aws/ClientBuilder.java
```diff
@@ -0,0 +1,8 @@
+package {{ package_name }};
+
+public class ClientBuilder {
+  // sample client builder - change ServiceSdkClient on the desired one
+  public static ServiceSdkClient getClient() {
```
surega21, line 5: Should this be commented out ? Otherwise it will not build right ?

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,131 @@
...
+        final ResourceHandlerRequest<ResourceModel> request,
+        final CallbackContext callbackContext,
+        final ProxyClient<ServiceSdkClient> proxyClient,
+        final Logger logger) {
+
+        this.logger = logger;
+
+        final ResourceModel model = request.getDesiredResourceState();
+
+        // TODO: Adjust Progress Chain according to your implementation
```
surega21, line 31: I think we should add more comments on explaining what is ProgressEvent

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,131 @@
...
+
+        return ProgressEvent.progress(model, callbackContext)
+            .then(progress -> checkForPreCreateResourceExistence(proxy, request, progress))
+            .then(progress ->
+                // If your service API throws 'ResourceAlreadyExistsException' for create requests then CreateHandler can return just proxy.initiate construction
+                proxy.initiate("Service-Name::{{ operation }}-Custom-Resource", proxyClient, model, callbackContext)
```
surega21, line 37: I was assuming this code-gen will be build-able so that the subsequent infrastructure can be set up as well. That would mean the actual implementation will be commented out with explanation on each API/helper methods defined. and meanwhile the handler will simply return a success event.

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
```diff
@@ -0,0 +1,71 @@
...
+
+        final ResourceModel model = request.getDesiredResourceState();
+
+        // TODO: Adjust Progress Chain according to your implementation
+
+        return proxy.initiate("Service-Name::Delete-Custom-Resource", proxyClient, model, callbackContext)
+            .request(Translator::translateToDeleteRequest) // construct a body of a request
+            .call(this::deleteResource) // make an api call
+            .stabilize(this::stabilizeOnDelete) // stabilize is describing the resource until it is in a certain status
+            .success(); // success if stabilized
```
aygold92, line 30: This call chain only represents one generic way to stabilize, and it needs to be explicitly noted as such (which is why I suggest commenting it out).  I would recommend using a real working example (like, take Cloudwatch logs or SES or some other simple resource) rather than fake code that doesn't work.

Also, I am confused by `"Service-Name::Delete-Custom-Resource"`.  I would have no idea what to put there

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    Boolean stabilizeOnDelete(
+        final ResourceModel model,
+        final ProxyClient<ServiceSdkClient> proxyClient) {
+        final AwsRequest readRequest = Translator.translateToReadRequest(model);
+        try {
+            proxy.injectCredentialsAndInvokeV2(readRequest, ClientBuilder.getClient()::executeReadRequest);
+            logger.log(String.format("%s [%s] is not yet deleted. Stabilization is in progress.", ResourceModel.TYPE_NAME, model.getPrimaryIdentifier()));
+            return false; // not stabilized
+        } catch(final ResourceNotFoundException e) {
```
aygold92, line 66: again, this is only one way to stabilize. It might be waiting for a certain status or some combination of other things.

python/rpdk/java/codegen.py
aygold92: also what are these other ones? I only see anything about guided_aws in here

ammokhov: this generates a simple class with four properties: 
```python
__dict__: 
    {
        default: '/default', 
        guided: '/guided', 
        default_code: '1', 
        guided_code: '2',
}
```
codes are to validate user inputs from the prompt and associate correct directory (`/default` or `/guided`) to generate code from the stubs


ln:21 is setting associated defaults. 
@surega21 `default` is a directory path to the standard stubs
@aygold92  this is used in order to specify paths and codes once and reuse across the class.

this function uses this class - `_prompt_for_codegen_model` (ln:90)
 |- `guided_aws` is the directory path for the new stubs, it is referenced via `CODEGEN.guided` ln:100
this `_is_aws_guided` checks if we are in guided mode (ln:108)

python/rpdk/java/templates/init/guided_aws/ClientBuilder.java
ammokhov: it wont build. previous commit has a buildable version but it's less explicit

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
ammokhov: since `ProgressEvent` is framework agnostic (used in both OOP and FP) then would it make sense to put in public docs and not in the codegen stubs?

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
ammokhov: good point, yea it's more like a guidelines at this point.

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,131 @@
...
+     * step in a discrete call/stabilize chain to ensure the entire resource is provisioned as intended.
+     * @param progressEvent
+     * @return
+     */
+    ProgressEvent<ResourceModel, CallbackContext> postCreateUpdate(
+        final ProgressEvent<ResourceModel, CallbackContext> progressEvent,
+        final ProxyClient<ServiceSdkClient> proxyClient) {
+        try {
+            proxyClient.injectCredentialsAndInvokeV2(Translator.translateToUpdateRequest(model), ClientBuilder.getClient()::executeUpdateResource);
+        } catch (final CfnResourceConflictException e) {
```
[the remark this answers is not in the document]
ammokhov: I would say that nothing here is necessary. it's just the most common pattern which might vary significantly based off resource complexity. Like in this example we tried to outline how to use `.request`-`.call` chain, `.stabilize`, and `.then`. Exceptions are just an example of how to handle it. The more generic the boiler plate the less clear it gets I suppose

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
ammokhov: well it seems to be a bit overkill bc the open sourced repo doesnt use this pattern but boiler plate will? if so then we should refactor existing code of cloudwatch. But I like the idea

python/rpdk/java/templates/init/default/HandlerWrapper.java
```diff
@@ -0,0 +1,152 @@
+// This is a generated file. Modifications will be overwritten.
```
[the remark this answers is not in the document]
ammokhov: you're right, will change that

.pylintrc
```diff
@@ -12,6 +12,8 @@ disable=
     bad-continuation,           # clashes with black
     too-few-public-methods,     # triggers when inheriting
     ungrouped-imports,          # clashes with isort
+    no-self-use,                # clashes with E0213
+    duplicate-code,             # duplicate test code - can't be ignored inline (https://github.com/PyCQA/pylint/issues/214)
```
jotompki, line 16: I'd rather disable both of these on a function/line level than disable them from the whole project. I think I understand the second issue with the second one but where are these occurring in the PR?

python/rpdk/java/codegen.py
```diff
@@ -10,12 +11,30 @@
...
 
 from .resolver import translate_type
-from .utils import safe_reserved, validate_namespace
+from .utils import safe_reserved, validate_codegen_model, validate_namespace
 
 LOG = logging.getLogger(__name__)
 
 OPERATIONS = ("Create", "Read", "Update", "Delete", "List")
 EXECUTABLE = "cfn"
+AWSCODEGEN = namedtuple("AWSCODEGEN", "default guided default_code guided_code")
```
jotompki, line 20: what's the difference in values here between default and default_code and guided/guided_code? seem redundant

python/rpdk/java/codegen.py
```diff
@@ -66,19 +86,57 @@ def _prompt_for_namespace(self, project):
...
         self.package_name = ".".join(self.namespace)
 
-    def init(self, project):
-        LOG.debug("Init started")
+    @staticmethod
+    def _prompt_for_codegen_model(project):
+        prompt = "Choose codegen model - 1 (default) or 2 (guided-aws): "
 
-        self._prompt_for_namespace(project)
+        codegen_model = input_with_validation(
```
jotompki, line 93: any reason we just cant return the value that we need the settings key to be set to here:

```
        project.settings["codegen_template_path"] = input_with_validation(...)
```

python/rpdk/java/codegen.py
```diff
@@ -10,12 +11,30 @@
...
 
 LOG = logging.getLogger(__name__)
 
 OPERATIONS = ("Create", "Read", "Update", "Delete", "List")
 EXECUTABLE = "cfn"
+AWSCODEGEN = namedtuple("AWSCODEGEN", "default guided default_code guided_code")
+CODEGEN = AWSCODEGEN("default", "guided_aws", "1", "2")
+
+
+def logdebug(func: object):
```
jotompki, line 24: nice

python/rpdk/java/codegen.py
```diff
@@ -128,69 +223,63 @@ def init(self, project):
...
-        LOG.debug("Init complete")
+    @logdebug
+    def init_default(self, project, src, tst):
+        """Writing handlers and tests"""
+        self.init_handlers(project, src, tst)
+
+    @logdebug
+    def init_guided_aws(self, project, src, tst):
+        """Writing handlers and tests"""
+        self.init_handlers(project, src, tst)
```
jotompki, line 245: This could really be part of `init_shared` or bubbled up to avoid the duplication. Any reason not to? Is this what was causing the pylint error?

tests/test_codegen_guided_aws.py
```diff
@@ -0,0 +1,239 @@
...
+        return "2"
+    return ""
+
+
+def test_java_language_plugin_module_is_set():
+    plugin = JavaLanguagePlugin()
+    assert plugin.MODULE_NAME
+
+
+def test_initialize(project):
```
jotompki, line 42: does this test anything specific to the guided codegen or is just covering the codepath for coverage?

tests/test_codegen_guided_aws.py
```diff
@@ -0,0 +1,239 @@
...
+    handler_properties = template["Resources"]["TypeFunction"]["Properties"]
+
+    code_uri = "./target/{}-handler-1.0-SNAPSHOT.jar".format(project.hypenated_name)
+    assert handler_properties["CodeUri"] == code_uri
+    handler = "{}.HandlerWrapper::handleRequest".format(expected_group_id)
+    assert handler_properties["Handler"] == handler
+    assert handler_properties["Runtime"] == project._plugin.RUNTIME
+
+
+def test_generate(project):
```
jotompki, line 63: same here

tests/test_codegen_guided_aws.py
```diff
@@ -0,0 +1,239 @@
...
+    target.mkdir(exist_ok=True)
+    jar_paths = []
+    for i in range(count):
+        jar_path = target / "{}-{}.0-SNAPSHOT.jar".format(project.hypenated_name, i)
+        jar_path.touch()
+        jar_paths.append(jar_path)
+    return jar_paths
+
+
+def test__find_jar_zero(project):
```
jotompki, line 99: I don't think we need a lot of these tests they are duplicates of what are in: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/tests/test_codegen.py#L94

Nothing in init/generate is newly tested in these.

python/rpdk/java/templates/init/guided_aws/BaseHandlerStd.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  protected abstract ProgressEvent<ResourceModel, CallbackContext> handleRequest(
+    final AmazonWebServicesClientProxy proxy,
+    final ResourceHandlerRequest<ResourceModel> request,
+    final CallbackContext callbackContext,
+    final ProxyClient<SdkClient> proxyClient,
+    final Logger logger);
+
+  // Placeholder for the functionality that could be shared across Create/Read/Update/Delete/List Handlers
```
rjlohan, line 33: Maybe this comment can just be a class comment to describe the purpose of this type? It's likely to get overlooked here and just appear as junk in handlers.

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,142 @@
+package {{ package_name }};
+
+// TODO: replace all usage of SdkClient with your service client type, e.g; YourServiceAsyncClient
+//import com.amzn.my.resource.ClientBuilder.SdkClient;
```
rjlohan, line 4: Should the import line be something like;

`import software.amazon.awssdk.services.yourservice.YourServiceAsyncClient;`

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,142 @@
...
+import software.amazon.cloudformation.exceptions.CfnNotFoundException;
+import software.amazon.cloudformation.proxy.AmazonWebServicesClientProxy;
+import software.amazon.cloudformation.proxy.Logger;
+import software.amazon.cloudformation.proxy.ProgressEvent;
+import software.amazon.cloudformation.proxy.ProxyClient;
+import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
+
+public class {{ operation }}Handler extends BaseHandlerStd {
+    // CallGraph value helps tracking the execution flow within the callback
+    // TODO: change this if you care, or leave it
```
rjlohan, line 22: I rethought this and maybe let's move this back as an inline string with the codegenerated value and no commentary. Most teams will just ignore this. Docs can explain it? Just means one less thing to have TODO.

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
rjlohan: We could put JavaDocs on the `ProgressEvent` class if it's not there, but otherwise don't put the comments in here I don't think.

python/rpdk/java/templates/init/guided_aws/StubHandler.java
```diff
@@ -0,0 +1,31 @@
+package {{ package_name }};
```
rjlohan, line 1: I don't think this file is used anymore since we have explicit stubs for each action?

python/rpdk/java/templates/init/guided_aws/StubHandlerTest.java
```diff
@@ -0,0 +1,58 @@
+package {{ package_name }};
```
rjlohan, line 1: Ditto this file - not needed?

python/rpdk/java/templates/init/guided_aws/StubListHandler.java
```diff
@@ -0,0 +1,31 @@
...
+    @Override
+    public ProgressEvent<{{ pojo_name }}, CallbackContext> handleRequest(
+        final AmazonWebServicesClientProxy proxy,
+        final ResourceHandlerRequest<{{ pojo_name }}> request,
+        final CallbackContext callbackContext,
+        final Logger logger) {
+
+        final List<ResourceModel> models = new ArrayList<>();
+
+        // TODO : put your code here
```
rjlohan, line 23: Should we do a bit more here to demonstrate a sensible call pattern with `NextToken` handling for example?

python/rpdk/java/templates/init/guided_aws/Translator.java
```diff
@@ -0,0 +1,121 @@
...
+   */
+  static List<ResourceModel> translateFromListRequest(final AwsResponse awsResponse) {
+    // e.g. e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/Translator.java#L81
+    return streamOfOrEmpty(Lists.newArrayList())
+        .map(resource -> ResourceModel.builder()
+            //.primaryIdentifier(resource.primaryIdentifier())
+            .build())
+        .collect(Collectors.toList());
+  }
+
```
rjlohan, line 114: There's 2 blank lines everywhere in this file?

python/rpdk/java/templates/init/guided_aws/StubListHandler.java
ammokhov: sometimes in list handler pattern wont work as callback traces the response values, which might exceed the size limit

tests/test_codegen_guided_aws.py
ammokhov: @rjlohan

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,138 @@
...
+                    .progress())
+
+            // STEP 3 [TODO: describe call/chain to return the resource model]
+            .then(progress -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
+    }
+
+    /**
+     * If your service API does not distinguish duplicate create requests against some identifier (e.g; resource Name)
+     * and instead returns a 200 even though a resource already exists, you must first check if the resource exists here
+     * NOTE: If your service API throws 'ResourceAlreadyExistsException' for create requests this method is not necessary
```
aygold92, line 65: i would include the word "idempotent" in here somewhere. 

"If your service API is not idempotent, meaning it does not distinguish..." or something like that

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,138 @@
...
+     */
+    private AwsResponse createResource(
+        final AwsRequest awsRequest,
+        final ProxyClient<SdkClient> proxyClient) {
+        AwsResponse awsResponse = null;
+        try {
+            // TODO: add custom create resource logic
+            // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/CreateHandler.java#L37-L43
+
+        // Proxy framework handles the majority of standardized aws exceptions
```
aygold92, line 102: such as? or link to docs?

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,138 @@
...
+            throw new CfnGeneralServiceException(ResourceModel.TYPE_NAME, e);
+        }
+
+        logger.log(String.format("%s successfully created.", ResourceModel.TYPE_NAME));
+        return awsResponse;
+    }
+
+    /**
+     * If your service does not provide strong consistency, you will need to account for eventual consistency issues
+     * ensure you stabilize your Create request so that a subsequent Read will successfully describe the resource state that was provisioned
```
aygold92, line 114: again, "strong consistency" is only one example of stabilization.  link to handler contract

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,138 @@
...
+     * @param proxyClient the aws service client to make the call
+     * @return awsResponse create resource response
+     */
+    private AwsResponse createResource(
+        final AwsRequest awsRequest,
+        final ProxyClient<SdkClient> proxyClient) {
+        AwsResponse awsResponse = null;
+        try {
+            // TODO: add custom create resource logic
+            // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/CreateHandler.java#L37-L43
```
aygold92, line 100: I wouldn't call this "custom create resource logic"... all we are doing is calling create.  Surely there is a better example?

This is a good example of an API that throws AlreadyExists exception though. I would link here in `step 1 [check if resource already exists]`

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,138 @@
...
+                proxy.initiate("{{ call_graph }}::{{ operation }}", proxyClient, model, callbackContext)
+
+                    // STEP 2.1 [TODO: construct a body of a request]
+                    .request(Translator::translateToCreateRequest)
+
+                    // STEP 2.2 [TODO: make an api call]
+                    .call(this::createResource)
+
+                    // STEP 2.3 [TODO: stabilize is describing the resource until it is in a certain status]
+                    .stabilize(this::stabilizedOnCreate)
```
aygold92, line 55: can we add optional postCreate as well? again, this is described in the handler contract

"""
Stabilization is the process of waiting for a resource to be in a particular state.  For Create and Update handlers, desired-state stabilization is satisfied when all properties specified in the request are applied as requested.  This is verified by calling the Read handler.  Note that reaching the desired-state is mandatory for all handlers before returning SUCCESS.

In many cases, the desired-state is reached immediately upon completion of the Create/Update API call.  However, in some cases, multiple API calls and/or wait periods may be required in order to reach this state.
"""

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,138 @@
...
+        final AwsRequest awsRequest,
+        final AwsResponse awsResponse,
+        final ProxyClient<SdkClient> proxyClient,
+        final ResourceModel model,
+        final CallbackContext callbackContext) {
+
+        // TODO: add custom stabilization logic
+
+        // hint: for some resources just describing current status would be enough
+        // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/ReadHandler.java#L40-L46
```
aygold92, line 132: why is this a link to the read handler? actually, it looks like log group does not stabilize at all.

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
```diff
@@ -0,0 +1,133 @@
...
+        final ProgressEvent<ResourceModel, CallbackContext> progressEvent) {
+        final ResourceModel model = progressEvent.getResourceModel();
+        final CallbackContext callbackContext = progressEvent.getCallbackContext();
+        try {
+            // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/ReadHandler.java#L40-L46
+            new ReadHandler().handleRequest(proxy, request, callbackContext, logger);
+            return ProgressEvent.progress(model, callbackContext);
+        } catch (CfnNotFoundException e) {
+            logger.log(model.getPrimaryIdentifier() + " does not exist; creating the resource.");
+            return ProgressEvent.success(model, callbackContext);
```
aygold92, line 78: the handler contract says we should Fail with NotFound exception in this case

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
```diff
@@ -0,0 +1,133 @@
...
+                // STEP 2.0 [initialize a proxy context]
+                proxy.initiate("{{ call_graph }}::{{ operation }}", proxyClient, model, callbackContext)
+
+                    // STEP 2.1 [TODO: construct a body of a request]
+                    .request(Translator::translateToDeleteRequest)
+
+                    // STEP 2.2 [TODO: make an api call]
+                    .call(this::deleteResource)
+
+                    // STEP 2.3 [TODO: stabilize is describing the resource until it is in a certain status]
```
aygold92, line 52: same comments as for create handler

python/rpdk/java/templates/init/guided_aws/StubReadHandler.java
```diff
@@ -0,0 +1,82 @@
...
+        final ProxyClient<SdkClient> proxyClient,
+        final Logger logger) {
+
+        this.logger = logger;
+
+        final ResourceModel model = request.getDesiredResourceState();
+
+        // TODO: Adjust Progress Chain according to your implementation
+
+        // STEP 1 [initialize a proxy context]
```
aygold92, line 34: this call chain paradigm doesn't seem necessary for Read

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -0,0 +1,129 @@
...
+
+            // STEP 4 [TODO: stabilize is describing the resource until it is in a certain status]
+            .stabilize(this::stabilizedOnUpdate)
+            .progress()
+
+            // STEP 5 [TODO: post stabilization update]
+            .then(progress -> postUpdate(progress, proxyClient))
+
+            // STEP 6 [TODO: describe call/chain to return the resource model]
+            .then(progress -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
```
aygold92, line 50: I think this needs to be reworked.  On create, the first API call is special (in 99% of cases) because it actually creates the resource.  Therefore, it makes sense to have a "create" and a "postCreate".  

However, on update there is no real distinguishment between the first API call (the "update") and the rest (the "postUpdate"), and there may be any number of API calls and stabilizations required, and not necessarily in the order shown here.

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
```diff
@@ -0,0 +1,133 @@
...
+            .then(progress ->
+                // If your service API throws 'ResourceNotFoundException' for delete requests then DeleteHandler can return just proxy.initiate construction
+                // STEP 2.0 [initialize a proxy context]
+                proxy.initiate("{{ call_graph }}::{{ operation }}", proxyClient, model, callbackContext)
+
+                    // STEP 2.1 [TODO: construct a body of a request]
+                    .request(Translator::translateToDeleteRequest)
+
+                    // STEP 2.2 [TODO: make an api call]
+                    .call(this::deleteResource)
```
aygold92, line 50: in many cases, there are some predelete actions that need to be done.  For example, see AWS::ECS::Service.  I would include an optional step here.

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
aygold92: i see this is actually described pretty well below in the comments for postUpdate (side note: I would again link to the handler contract for desired-state stabilization), but the call chain setup does not really make sense

python/rpdk/java/templates/init/guided_aws/Translator.java
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Translates resource objects from sdk into a resource model (primary identifier only)
+   * @param awsResponse the aws service describe resource response
+   * @return list of resource models
+   */
+  static List<ResourceModel> translateFromListRequest(final AwsResponse awsResponse) {
+    // e.g. e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/Translator.java#L81
+    return streamOfOrEmpty(Lists.newArrayList())
+        .map(resource -> ResourceModel.builder()
```
aygold92, line 103: I thought List is meant to return a list of primaryIdentifiers, not a list of resource models. perhaps this has not been updated across all the packages/interfaces...

python/rpdk/java/templates/init/guided_aws/Translator.java
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * Translates resource object from sdk into a resource model
+   * @param awsResponse the aws service describe resource response
+   * @return model resource model
+   */
+  static ResourceModel translateFromReadResponse(final AwsResponse awsResponse) {
+    // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/Translator.java#L58
+    return ResourceModel.builder()
+        //.primaryIdentifier(response.primaryIdentifier())
```
aygold92, line 54: there isn't actually a property called `primaryIdentifier` that is meant to be set.  I would be confused by this.  I think you should just remove this line

python/rpdk/java/templates/init/guided_aws/Translator.java
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Translates resource objects from sdk into a resource model (primary identifier only)
+   * @param awsResponse the aws service describe resource response
+   * @return list of resource models
+   */
+  static List<ResourceModel> translateFromListRequest(final AwsResponse awsResponse) {
+    // e.g. e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/Translator.java#L81
+    return streamOfOrEmpty(Lists.newArrayList())
+        .map(resource -> ResourceModel.builder()
+            //.primaryIdentifier(resource.primaryIdentifier())
```
aygold92, line 104: same here remove this line

python/rpdk/java/templates/init/guided_aws/Translator.java
rjlohan: We don't know what the properties are, so this is just suggestive. Can remove though.

python/rpdk/java/templates/init/guided_aws/Translator.java
rjlohan: But this is suggestive of how to return a list of models containing only the `primaryIdentifier`.

.travis.yml
```diff
@@ -40,6 +40,6 @@ jobs:
         - DIR=$(mktemp -d)
         - cd "$DIR"
         - ls -la
-        - printf "AWS::Foo::Bar\n\n" | cfn init -vv
+        - printf "AWS::Foo::Bar\n\n2" | cfn init -vv
```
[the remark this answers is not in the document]
ammokhov: this is a `guided_code` so the build can verify `aws guided` codegen model

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    private ProgressEvent<ResourceModel, CallbackContext> postUpdate(
+        final ProgressEvent<ResourceModel, CallbackContext> progressEvent,
+        final ProxyClient<SdkClient> proxyClient) {
+
+        final ResourceModel model = progressEvent.getResourceModel();
+        final CallbackContext callbackContext = progressEvent.getCallbackContext();
+        try {
+            // TODO: put your post update resource code here
+            // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/UpdateHandler.java#L69-L74
```
aygold92, line 121: broken link.  Also, let's not link to example code that isn't using the call chain when we explicitly want them to use a call chain

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -0,0 +1,133 @@
...
+            .call(this::updateResource)
+
+            // STEP 4 [TODO: stabilize step is not necessarily required but typically involves describing the resource until it is in a certain status, though it can take many forms]
+            // for more information -> https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html#resource-type-test-contract-communication
+            .stabilize(this::stabilizedOnUpdate)
+            .progress()
+
+            // If your resource is provisioned through multiple API calls, then the following pattern is required (and might take as many postUpdate callbacks as necessary)
+            // STEP 5 [TODO: post stabilization update]
+            .then(progress -> postUpdate(progress, proxyClient))
```
aygold92, line 49: I like this description 👍  

However, as noted before can we use a different function name besides `postUpdate()`?  It doesn't really make sense since it's all 1 update.  It's a tough generic name, I'd rather call it like `firstUpdateCall()` `secondUpdateCall()`, etc.

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -0,0 +1,133 @@
...
+        // STEP 1 [initialize a proxy context]
+        return proxy.initiate("{{ call_graph }}::{{ operation }}", proxyClient, model, callbackContext)
+
+            // STEP 2 [TODO: construct a body of a request]
+            .request(Translator::translateToUpdateRequest)
+
+            // STEP 3 [TODO: make an api call]
+            .call(this::updateResource)
+
+            // STEP 4 [TODO: stabilize step is not necessarily required but typically involves describing the resource until it is in a certain status, though it can take many forms]
```
aygold92, line 42: can we note that stabilization may or may not be needed after each API call?  Currently it seems like this is the proper order "first do an update, then stabilize, then do rest of updates, then you're done"

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
aygold92: Also, I'm just realizing what's confusing about this... the `updateResource` method is inside of a call chain and so we just make the API call in that method.  However, the `postUpdate` method requires a new call chain inside of it.

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
```diff
@@ -0,0 +1,136 @@
...
+        final ResourceHandlerRequest<ResourceModel> request,
+        final ProgressEvent<ResourceModel, CallbackContext> progressEvent) {
+        final ResourceModel model = progressEvent.getResourceModel();
+        final CallbackContext callbackContext = progressEvent.getCallbackContext();
+        try {
+            // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/ReadHandler.java#L28-L33
+            new ReadHandler().handleRequest(proxy, request, callbackContext, logger);
+            return ProgressEvent.progress(model, callbackContext);
+        } catch (CfnNotFoundException e) { // ResourceNotFoundException
+            logger.log(model.getPrimaryIdentifier() + " does not exist.");
```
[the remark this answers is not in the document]
ammokhov: not sure if it's needed. the check already happened. At that point we know that there is nothing to delete

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        final ResourceHandlerRequest<ResourceModel> request,
+        final CallbackContext callbackContext,
+        final ProxyClient<SdkClient> proxyClient,
+        final Logger logger) {
+
+        this.logger = logger;
+
+        final ResourceModel model = request.getDesiredResourceState();
+
+        // TODO: Adjust Progress Chain according to your implementation
```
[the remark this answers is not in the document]
ammokhov: the classes are already overloaded with information. I think at this point a developer needs to calibrate pattern and see what's available. This serves a specific example pretty well. If there will be a confusion on how to use a specific method then there will be examples (repos) and public doc

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
ammokhov: it depends on the complexity. ur proxy chain might fork depends on the template

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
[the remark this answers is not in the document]
aygold92: i can tell you that for debugging purposes it would be very helpful to have the requestId/message

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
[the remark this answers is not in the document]
aygold92: I'm not saying we need to document the entire framework here but a link to documentation would be very helpful

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            // TODO: put your delete resource code here
+            // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/commit/2077c92299aeb9a68ae8f4418b5e932b12a8b186#diff-82aedf3b1de1e4a4948229fa3cac03f0R21-R30
+
+            /*
+             * While the handler contract states that the handler must always return a progress event,
+             * you may throw any instance of BaseHandlerException (https://code.amazon.com/packages/AWSCloudFormationRPDKJavaPlugin/blobs/mainline/--/src/main/java/software/amazon/cloudformation/exceptions/BaseHandlerException.java),
+             * as the wrapper map it to a progress event. Each BaseHandlerException maps to a specific error code, and you should map service exceptions as closely as possible
+             * to more specific error codes (https://code.amazon.com/packages/AWSCloudFormationRPDKJavaPlugin/blobs/mainline/--/src/main/java/software/amazon/cloudformation/proxy/HandlerErrorCode.java)
+             */
```
aygold92, line 103: nit: this would make more sense in the catch block

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,176 @@
...
+     */
+    private ProgressEvent<ResourceModel, CallbackContext> postCreate(
+        final ProgressEvent<ResourceModel, CallbackContext> progressEvent,
+        final ProxyClient<SdkClient> proxyClient) {
+
+        final ResourceModel model = progressEvent.getResourceModel();
+        final CallbackContext callbackContext = progressEvent.getCallbackContext();
+        try {
+            // TODO: put your post creation resource update code here
+            // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/commit/2077c92299aeb9a68ae8f4418b5e932b12a8b186#diff-c959c290785791b96fa4442573750fcdR69-R74
```
aygold92, line 161: again, we're asking them explicitly to use a call chain pattern and then saying "here's an example of what I'm talking about" and that example doesn't use the call chain pattern...

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -0,0 +1,139 @@
...
+
+            // STEP 2 [TODO: construct a body of a request]
+            .translate(Translator::translateToUpdateRequest)
+
+            // STEP 3 [TODO: make an api call]
+            .call(this::updateResource)
+
+            // STEP 4 [TODO: stabilize step is not necessarily required but typically involves describing the resource until it is in a certain status, though it can take many forms]
+            // stabilization step may or may not be needed after each API call
+            // for more information -> https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html#resource-type-test-contract-communication
```
aygold92, line 44: again, remove the `#resource-type-test-contract-communication` from all the URLs because this will change

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -0,0 +1,139 @@
...
+        // TODO: Adjust Progress Chain according to your implementation
+
+        // STEP 1 [initialize a proxy context]
+        return proxy.initiate("{{ call_graph }}::{{ operation }}", proxyClient, model, callbackContext)
+
+            // STEP 2 [TODO: construct a body of a request]
+            .translate(Translator::translateToUpdateRequest)
+
+            // STEP 3 [TODO: make an api call]
+            .call(this::updateResource)
```
aygold92, line 40: I'm still confused why this first update has the call chain already set up but the second update doesn't have a call chain.  I just think it should be consistent.  So maybe instead:
```
ProgressEvent.progress(model, callbackContext)
.then(progress -> firstUpdate(proxy, request, progress))
.then(progress -> secondUpdate(proxy, request, progress))
.then(progress -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
```

Then, inside the firstUpdate and secondUpdate methods, you can already have the call chain setup with proxy.initiate.  Same for all handlers, including create/postCreate and delete

ammokhov: I dont think we should have different approaches for different handlers. The second update might not be necessary, so it's optional. We are just showing the way to use it, then a developer can modify as preferred.

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
ammokhov: removed

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
aygold92: I'm not suggesting different approaches for different handlers, i'm suggesting that we **don't have two different approaches in the same handler**.  First, you call out to a method which doesn't need a call chain, and then you call out to a method which DOES need a call chain (and then we link to an example which doesn't use the call chain)

.pylintrc
jotompki: where's the 'no-self-use' occurring? cant seem to find it looking at the PR

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -0,0 +1,139 @@
...
+
+            // STEP 4 [TODO: stabilize step is not necessarily required but typically involves describing the resource until it is in a certain status, though it can take many forms]
+            // stabilization step may or may not be needed after each API call
+            // for more information -> https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html
+            .stabilize(this::stabilizedOnFirstUpdate)
+            .progress()
+
+            // If your resource is provisioned through multiple API calls, then the following pattern is required (and might take as many postUpdate callbacks as necessary)
+            // STEP 5 [TODO: post stabilization update]
+            .then(progress -> secondUpdate(progress, proxyClient))
```
wbingli, line 50: I would suggest we make this secondUpdate as a comment. Not so many resource needs a second update. But if they do, they knows to add a new method.

python/rpdk/java/templates/init/guided_aws/ClientBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+// import software.amazon.awssdk.services.yourservice.YourServiceAsyncClient;
+// import software.amazon.cloudformation.LambdaWrapper;
+
+public class ClientBuilder {
+  /*
+  TODO: uncomment the following, replacing YourServiceAsyncClient with your service client name
+  It is recommended to use static HTTP client so less memory is consumed
+  e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/ClientBuilder.java#L9
+
+  private static class LazyHolder {
```
wbingli, line 14: Why we need an extra static class?  The SERVICE_CLIENT could be static for the `ClientBuilder` itself.

python/rpdk/java/templates/init/guided_aws/POJO.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class {{ model_name|uppercase_first_letter }} {
```
wbingli, line 18: Please also include the recent fix for the POJO:  https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/commit/ddfbcbf34fc04b7b4ca63656be67e2514092598a

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                    // STEP 2.2 [TODO: make an api call]
+                    .call(this::createResource)
+
+                    // STEP 2.3 [TODO: stabilize step is not necessarily required but typically involves describing the resource until it is in a certain status, though it can take many forms]
+                    // for more information -> https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html
+                    .stabilize(this::stabilizedOnCreate)
+                    .progress())
+
+            // STEP 3 [TODO: post create/stabilize update]
+            .then(progress -> postCreate(progress, proxyClient))
```
wbingli, line 61: nit:  Maybe comment the postCreate instead not too many resource do postCreate after stabilization.

wbingli:
LGTM. But keep an eye on the POJO.java for a [recent fix](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/commit/ddfbcbf34fc04b7b4ca63656be67e2514092598a)

.pylintrc
ammokhov: yep, removed. parametrized fixture instead in the test module so this change is not necessary

Issue #151, opened 2019-08-30
"integer" type in schema cannot be generated in POJO as a "long"


rjlohan:
The code generation for POJO uses `integer` for any `{ "type": "integer" }` in the schema. There is no way to make this a long instead without disabling codegen and manipulating the POJO directly. Would be good to either document the workaround or make it easier for developers to tweak the generated POJO somehow.

bkabrda:
Hey folks, is there any progress on this? We (maintainers of https://github.com/DataDog/datadog-cloudformation-resources) would very much like to see this. Right now we're marking the fields that need to be generated as `Long` with `number`, which generates them as `Double`. This unfortunately poses some problems, as explained in [1]. We have a user who posted a PR to instead use strings [2], which works better, but is still just a workaround for the inability to have long integers. For us, the solution proposed in #244 works fine (as long as "Physical ID" would just be a number, e.g. `123456789` and not end with `L`, e.g. `123456789L`, which would pose the same set of problems as reported in [1]).

Thanks!

[1] https://github.com/DataDog/datadog-cloudformation-resources/issues/80
[2] https://github.com/DataDog/datadog-cloudformation-resources/pull/81

jotompki:
Should be addressed by #328.  Will close once released.

jotompki:
Released in https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/releases/tag/v2.0.3-plugin

PR #244, opened 2020-04-21
Change Integer -> Long

jotompki:
*Issue #, if available:* #151 

*Description of changes:* This changes the codegen to generate a Long instead of Integer when generating code for schemas of type "integer". Going to open this as a draft PR and document any testing done to ensure this is a safe change to make and if not encorporate a version bump to signify the difference. 

### DIFF
--- a/python/rpdk/java/resolver.py
+++ b/python/rpdk/java/resolver.py
@@ -2,7 +2,7 @@ from rpdk.core.jsonutils.resolver import UNDEFINED, ContainerType
 
 PRIMITIVE_TYPES = {
     "string": "String",
-    "integer": "Integer",
+    "integer": "Long",
     "boolean": "Boolean",
     "number": "Double",
     UNDEFINED: "Object",


python/rpdk/java/resolver.py
```diff
@@ -2,7 +2,7 @@
 
 PRIMITIVE_TYPES = {
     "string": "String",
-    "integer": "Integer",
+    "integer": "Long",
```
wbingli, line 5: Is it backward compatible?  I'm afraid it will break existing resources have integer in their schema

jotompki: That's part of the reason why I opened it as a draft pr -- wanted to do some research to see if this was a safe change to make.

From looking it seems that if anyone is directly passing Integers from the pojo to an API request that requires an Integer, that would break. Might need a more sophisticated approach than just changing it here.

rjlohan: Int32 constraints can be achieved with use of `minimum` and `maximum` in the schema. If the actual downstream API calls are expecting a Java `Integer` in their payload, then they'll get a compile error on next build and it's a signal they'll need to update something. 

This doesn't break anything already published, so we should make this change before it gets too hairy. Suggest we could also add tests around this to signal there's a change in the POJO, but I'm not sure what those would look like.

wbingli: Could we introduce something in the `.rpdk-config` that we can prefer to use Long instead of Integer?  It won't break existing customer even in the build time.   The java plugin only need to read the `.rpdk-config` for a config and decide to use Integer or Long, but by default it's Integer.

bkabrda: FWIW I'd appreciate the solution proposed by @rjlohan, as it would allow us to distinguish between individual `Integer`s/`Long`s explicitly, unlike the other solutions which would require us to have *everything* `Long` (to be fair, having everything `Long` is still better, but would require us to do a conversion `Long`->`Integer` in quite a few places).

rjlohan: If we go with using constraints as the determination for `int32` vs `long`/`int64` we probably just need to have a check on whether the `maximum` is `< 2147483647` (using whatever const Python3 has) to determine if we codegen as `int32`. Else we use `long`. Another alternative would be to introduce a new `long` type to our metaschema to be explicit.

bkabrda: Yeah, I'm ok with determining this based on the `maximum`. If you want to go with introducing new type, I'd much rather see something along the line of OpenAPI spec, which has `type: integer` and it provides additional information in another property, e.g. `format: int32` or `format: int64`: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#dataTypeFormat

rjlohan: The `format` approach is a better idea, let's do that.

jotompki:
Superseded by #328

PR #253, opened 2020-04-30
Version bump to 1.0.4

wbingli:
*Issue #, if available:* N/A

*Description of changes:*. Release the latest change.


wbingli:
The integration test failure is the reason why we need to release this new version.  Need to update the default code gen version to 1.0.4 so the integ test will use the version from this repo instead of pulling 1.0.3 from remote maven repository.

PR #255, opened 2020-05-05
added compression to handler response stream

ammokhov:
*Issue #, if available:*

*Description of changes:*

added response compression. new format -> 
```
{
    '__COMPRESSED__': gziped_base64_compressed_string,
    '__COMPRESSION_METHOD__': 'gzip_base64'
}
```

### DIFF
--- a/src/main/java/software/amazon/cloudformation/resource/Serializer.java
+++ b/src/main/java/software/amazon/cloudformation/resource/Serializer.java
@@ -38,7 +38,7 @@ public class Serializer {
 
     public static final String COMPRESSED = "__COMPRESSED__";
     private static final String COMPRESSION_METHOD = "__COMPRESSION_METHOD__";
-    private static final String COMPRESSION = "gzip_base64";
+    private static final String COMPRESSION_GZIP_BASE64 = "gzip_base64";
     private static final ObjectMapper OBJECT_MAPPER;
     private static final ObjectMapper STRICT_OBJECT_MAPPER;
     private static final TypeReference<Map<String, Object>> MAP_TYPE_REFERENCE = new TypeReference<Map<String, Object>>() {
@@ -96,7 +96,7 @@ public class Serializer {
                 gzip.write(modelInput.getBytes(StandardCharsets.UTF_8));
             }
             map.put(COMPRESSED, Base64.encodeBase64String(byteArrayOutputStream.toByteArray()));
-            map.put(COMPRESSION_METHOD, COMPRESSION);
+            map.put(COMPRESSION_METHOD, COMPRESSION_GZIP_BASE64);
         }
         return OBJECT_MAPPER.writeValueAsString(map);
     }
--- a/src/main/java/software/amazon/cloudformation/resource/Serializer.java
+++ b/src/main/java/software/amazon/cloudformation/resource/Serializer.java
@@ -38,6 +38,7 @@ public class Serializer {
 
     public static final String COMPRESSED = "__COMPRESSED__";
     private static final String COMPRESSION_METHOD = "__COMPRESSION_METHOD__";
+    private static final String COMPRESSION = "gzip_base64";
     private static final ObjectMapper OBJECT_MAPPER;
     private static final ObjectMapper STRICT_OBJECT_MAPPER;
     private static final TypeReference<Map<String, Object>> MAP_TYPE_REFERENCE = new TypeReference<Map<String, Object>>() {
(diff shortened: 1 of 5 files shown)


src/main/java/software/amazon/cloudformation/resource/Serializer.java
```diff
@@ -76,10 +87,37 @@
         return OBJECT_MAPPER.writeValueAsString(modelObject);
     }
 
+    public <T> String compress(final String modelInput) throws IOException {
+        final Map<String, String> map = new HashMap<>();
+        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+        GZIPOutputStream gzip = new GZIPOutputStream(byteArrayOutputStream);
+        gzip.write(modelInput.getBytes(StandardCharsets.UTF_8));
+        gzip.close();
+        map.put(COMPRESSED, Base64.encodeBase64String(byteArrayOutputStream.toByteArray()));
```
wbingli, line 96: How about put additional metadata as how it is compressed?  like `"__COMPRESSED_METHOD__": "gzip_base64"`

src/main/java/software/amazon/cloudformation/resource/Serializer.java
```diff
@@ -76,10 +87,37 @@
...
+    public String decompress(final String s) {
+        try {
+            final Map<String, String> map = deserialize(s, MAP_TYPE_REFERENCE);
+
+            if (!map.containsKey(COMPRESSED)) {
+                return s;
+            }
+
+            final byte[] bytes = Base64.decodeBase64(map.get(COMPRESSED));
+            ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(bytes);
```
wbingli, line 113: Same to use try-with-resource for InputStream.

src/main/java/software/amazon/cloudformation/resource/Serializer.java
```diff
@@ -76,10 +87,37 @@
...
+            GZIPInputStream gzipInputStream = new GZIPInputStream(byteArrayInputStream);
+            return new String(IOUtils.toByteArray(gzipInputStream), StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            return s;
```
wbingli, line 117: I don't see a reason why it would fail. But as least add logging here for the exception.

src/main/java/software/amazon/cloudformation/resource/Serializer.java
```diff
@@ -76,10 +88,48 @@
         return OBJECT_MAPPER.writeValueAsString(modelObject);
     }
 
+    /*
+     * public <T> String compress(final String modelInput) throws IOException {
```
wbingli, line 92: comment out code?

src/main/java/software/amazon/cloudformation/resource/Serializer.java
```diff
@@ -76,10 +88,48 @@
...
+     * gzip.write(modelInput.getBytes(StandardCharsets.UTF_8)); gzip.close();
+     * map.put(COMPRESSED,
+     * Base64.encodeBase64String(byteArrayOutputStream.toByteArray()));
+     * map.put(COMPRESSION_METHOD, "gzip_base64"); return
+     * OBJECT_MAPPER.writeValueAsString(map); }
+     */
+
+    public <T> String compress(final String modelInput) throws IOException {
+        final Map<String, String> map = new HashMap<>();
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
```
wbingli, line 105: I think you can do with one try

```
try (
ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
GZIPOutputStream gzip = new GZIPOutputStream(byteArrayOutputStream);
) {
...
}
```

src/main/java/software/amazon/cloudformation/resource/Serializer.java
```diff
@@ -76,10 +88,48 @@
...
+            try (GZIPOutputStream gzip = new GZIPOutputStream(byteArrayOutputStream)) {
+                gzip.write(modelInput.getBytes(StandardCharsets.UTF_8));
+            }
+            map.put(COMPRESSED, Base64.encodeBase64String(byteArrayOutputStream.toByteArray()));
+            map.put(COMPRESSION_METHOD, "gzip_base64");
```
wbingli, line 110: `gzip_base64` better to put as constant string.

src/main/java/software/amazon/cloudformation/resource/Serializer.java
wbingli: Seems the two `try` is right as it needs to finish first gzip output stream then starts the second one.

src/main/java/software/amazon/cloudformation/resource/Serializer.java
```diff
@@ -22,14 +23,26 @@
...
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
+import org.apache.commons.codec.binary.Base64;
 import software.amazon.cloudformation.proxy.aws.AWSServiceSerdeModule;
 
 public class Serializer {
 
+    public static final String COMPRESSED = "__COMPRESSED__";
+    private static final String COMPRESSION_METHOD = "__COMPRESSION_METHOD__";
+    private static final String COMPRESSION = "gzip_base64";
```
wbingli, line 41: nit: better rename?  `COMPRESSION_GZIP_BASED64`?

PR #259, opened 2020-05-07
Add proxy methods to support invocation for inputstream/bytes

xiwhuang:
### What are we changing?
* Added two proxy methods to extend `ResponseInputStream<AwsResponse>` and `ResponseBytes<AwsResponse>`

### Why are we changing?
* A few uncommon AWS APIs like `S3::GetObject` requires AWS Credentials Provider to authenticate the API requests

### What happens if we do not change?
* Customers have to mimic a dummy request to extract out the credentials. for example, [StackSet needs to validate the content in S3](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/6/files/1e7ca3fc2d9223b7ac244ca08258991b4216e2b8#diff-3f62f9b7133240c794680f6a69c6b645)

### How did we test?
* Manually tested the changes locally.
* Associated Unit tests pass.
* Create/Update/Delete succeed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,6 +155,10 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>apache-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+        </dependency>
 
         <!-- https://mvnrepository.com/artifact/commons-io/commons-io -->
         <dependency>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -155,6 +155,10 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>apache-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
```
jotompki, line 160: scope this for testing if it is only used in unit tests

jotompki:
```
<scope>test</scope>
```

xiwhuang: Sure.

Issue #243, opened 2020-04-17
Cfn error code exceptions do not use error messages from the cause exception, omitting service level error messages and request Ids


aygold92:
We created exceptions to represent the different error codes, specifically:

* CfnAccessDeniedException
* CfnAlreadyExistsException
* CfnGeneralServiceException
* CfnInternalFailureException
* CfnInvalidCredentialsException
* CfnInvalidRequestException
* CfnNetworkFailureException
* CfnNotFoundException
* CfnNotStabilizedException
* CfnNotUpdatableException
* CfnResourceConflictException
* CfnServiceInternalErrorException
* CfnServiceLimitExceededException
* CfnThrottlingException

The idea is that the handler would throw this exception, and the java plugin would convert that to a ProgressEvent with a FAILED status.  However, these exceptions do not take error messages, which is required when there is a FAILED ProgressEvent.  

Also, these exceptions should extend a common `BaseErrorCodeException` or somehting like that, so we can easily catch this.  Handlers should not be throwing any other type of exception from their code.

Lastly, NotFound and AlreadyExists have duplicate exceptions, and I'm not sure what they are for:
* ResourceNotFoundException and CfnNotFoundException
* ResourceAlreadyExistsException and CfnAlreadyExistsException

aygold92:
After doing a bit of digging it looks like we actually should use exception messages https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/ddfbcbf34fc04b7b4ca63656be67e2514092598a/src/main/java/software/amazon/cloudformation/proxy/ExceptionMessages.java#L17, but the constructors are pretty confusing and these messages never include AWS requestIds.

aygold92:
it also looks we use BaseHandlerException which includes the ErrorCode https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/ddfbcbf34fc04b7b4ca63656be67e2514092598a/src/main/java/software/amazon/cloudformation/exceptions/BaseHandlerException.java#L20

aygold92:
it looks like really what the work is:
1. remove `TerminalException` https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/exceptions/TerminalException.java, which is not used anywhere and is confusing, 
1. remove `ResourceAlreadyExistsException` and `ResourceNotFoundException` which are duplicates
[...]

PR #264, opened 2020-05-15
Fix base exception message

wbingli:
*Issue #, if available:*  #243 

*Description of changes:*. Fix the error message issue for BaseHandlerException. 

Cloudformation service doesn't like empty message exception, use the cause exception as default exception message.

### DIFF
--- a/src/main/java/software/amazon/cloudformation/exceptions/BaseHandlerException.java
+++ b/src/main/java/software/amazon/cloudformation/exceptions/BaseHandlerException.java
@@ -26,7 +26,7 @@ public abstract class BaseHandlerException extends RuntimeException {
 
     protected BaseHandlerException(final Throwable cause,
                                    final HandlerErrorCode errorCode) {
-        super(null, cause);
+        super(cause.getMessage(), cause);
         this.errorCode = errorCode;
     }
 
@@ -34,6 +34,12 @@ public abstract class BaseHandlerException extends RuntimeException {
                                    final Throwable cause,
                                    final HandlerErrorCode errorCode) {
         super(message, cause);
+        this.errorCode = errorCode;
+    }
+
+    protected BaseHandlerException(final String message,
+                                   final HandlerErrorCode errorCode) {
+        super(message);
         this.errorCode = errorCode;
     }
 }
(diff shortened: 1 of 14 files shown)


miparnisari:
So this is why this code

```java
try {
            awsResponse = proxyClient.injectCredentialsAndInvokeV2(awsRequest, proxyClient.client()::deleteMetricFilter);
        } catch (ResourceNotFoundException e) {
            logger.log("Resource does not exist and could not be deleted.");
            throw new CfnNotFoundException(e);
        }
```

causes this to be printed in the handler logs...

```

  | 2020-05-15T15:00:53.943-07:00 | Resource does not exist and could not be deleted.
  | 2020-05-15T15:00:54.011-07:00 | null in a DELETE action on a AWS::Logs::MetricFilter: software.amazon.cloudformation.exceptions.CfnNotFoundExceptionsoftware.amazon.cloudformation.exceptions.CfnNotFoundException	at software.amazon.logs.metricfilter.DeleteHandler.deleteResource(DeleteHandler.java:51)

File: src/main/java/software/amazon/cloudformation/exceptions/BaseHandlerException.java [hunk 1]
```diff
@@ -26,7 +26,7 @@
 
     protected BaseHandlerException(final Throwable cause,
                                    final HandlerErrorCode errorCode) {
-        super(null, cause);
+        super(cause.getMessage(), cause);
```
aygold92, line 29: I think you can just call `super(cause)`

src/main/java/software/amazon/cloudformation/exceptions/CfnInvalidCredentialsException.java
```diff
@@ -22,10 +22,10 @@
     private static final HandlerErrorCode ERROR_CODE = HandlerErrorCode.InvalidCredentials;
 
     public CfnInvalidCredentialsException() {
-        this(null);
+        super(ERROR_CODE.getMessage(), ERROR_CODE);
     }
 
     public CfnInvalidCredentialsException(final Throwable cause) {
```
aygold92, line 28: I don't see any reason to limit all of these exceptions to just having a `cause` constructor.  Can we include a `(cause)` and a `(message, cause)` for all of them?

aygold92: can we create an issue to resolve this then, if we're not going to include it in this PR?

wbingli: Tracked in #276

Issue #265, opened 2020-05-15
codegen - create/delete handler pre execution check 


ammokhov:
Currently `preExistanceCheck`/`preDeletionCheck` invoke `ReadHandler` which returns sub chain with call graph but it doesnt get converged into the main chain. This might lead to an unstable handler behavior on reinvocation - scenario:

Step 1: CREATE Invoked
Step 1.1: check for pre existance - not found, execution continues
Step 1.2: create a resource and time is up - lambda exited

Step 2: CREATE Reinvoked
Step 2.1: check for pre existance - resource found (created from previous execution) - throws already exist exception.

In this case Create Handler fails to reach `SUCCESS` status.

PR #266, opened 2020-05-15
fixed create/delete handler pre execution check

ammokhov:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/265

*Description of changes:*

- Refactoring crud handler stubs so it's less verbose and more one-liner. 
- `CreateHandler` and `DeleteHandler` have fixed pre execution checks which add on the main call graph - #265 

### DIFF
--- a/python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
+++ b/python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
@@ -3,19 +3,16 @@ package {{ package_name }};
 // TODO: replace all usage of SdkClient with your service client type, e.g; YourServiceAsyncClient
 // import software.amazon.awssdk.services.yourservice.YourServiceAsyncClient;
 
-import java.util.Objects;
-import software.amazon.awssdk.awscore.AwsRequest;
 import software.amazon.awssdk.awscore.AwsResponse;
 import software.amazon.awssdk.awscore.exception.AwsServiceException;
 import software.amazon.awssdk.core.SdkClient;
-import software.amazon.cloudformation.exceptions.CfnAlreadyExistsException;
 import software.amazon.cloudformation.exceptions.CfnGeneralServiceException;
-import software.amazon.cloudformation.exceptions.CfnNotFoundException;
 import software.amazon.cloudformation.proxy.AmazonWebServicesClientProxy;
 import software.amazon.cloudformation.proxy.Logger;
 import software.amazon.cloudformation.proxy.ProgressEvent;
 import software.amazon.cloudformation.proxy.ProxyClient;
 import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
+
 
 public class CreateHandler extends BaseHandlerStd {
     private Logger logger;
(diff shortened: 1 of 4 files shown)


python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -29,157 +26,92 @@
...
+                        final boolean stabilized = true;
+                        logger.log(String.format("%s [%s] has been stabilized.", ResourceModel.TYPE_NAME, model.getPrimaryIdentifier()));
+                        return stabilized;
+                    })
                     .progress()
                 )
 
-            // STEP 4 [TODO: describe call/chain to return the resource model]
+            // STEP 3 [TODO: describe call/chain to return the resource model]
             .then(progress -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
```
miparnisari, line 115: Why is this read call here okay but not the pre-create read?

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -29,157 +26,92 @@
...
+                    // STEP 1.3 [TODO: handle exception]
+                    .handleError((awsRequest, exception, client, model, context) -> {
+                        // TODO: uncomment when ready to implement
+                        // if (exception instanceof CfnNotFoundException)
+                        //     return ProgressEvent.progress(model, context);
+                        // throw exception;
+                        return ProgressEvent.progress(model, context);
+                    })
+                    .progress()
+            )
```
miparnisari, line 66: I don't really like this solution of embedding what is basically Read handler code in the Create handler. In my particular case (AWS::Logs::MetricFilter),the SDK API for Create and Update is the same, so I have to check for pre-existence in both handlers, which would mean having mostly the same code across three files.

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -29,157 +26,92 @@
...
+                        logger.log(String.format("%s has successfully been read.", ResourceModel.TYPE_NAME));
+                        return awsResponse;
+                    })
+
+                    // STEP 1.3 [TODO: handle exception]
+                    .handleError((awsRequest, exception, client, model, context) -> {
+                        // TODO: uncomment when ready to implement
+                        // if (exception instanceof CfnNotFoundException)
+                        //     return ProgressEvent.progress(model, context);
+                        // throw exception;
```
miparnisari, line 62: I tested throwing exceptions here, and they are not bubbled all the way up. The handler ends up returning a status code of FAILED but no exception is thrown.

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
ammokhov: read call at the end is to get a resource model with all the properties provisioned. pre check needs to be registered in the call chain as it gets reused. final step is not reused

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
ammokhov: u can create a single call chain in base handler then use it in both handlers

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
rjlohan: The intent is the Handler should return FAILED with a sane/consistent exception message that ends up in Stack Trace. I think there's a separate issue that needs to be fixed for that as it's not working reliably.

miparnisari: So which of these asserts should I use?

```java
assertThatThrownBy(() -> handler.handleRequest(proxy, request, new CallbackContext(), proxyClient, logger))
 .isInstanceOf(CfnAlreadyExistsException.class);
```
or

```java
response = handler.handleRequest(proxy, request, new CallbackContext(), proxyClient, logger);
assertThat(response).isNotNull();
assertThat(response.getStatus()).isEqualTo(OperationStatus.FAILED);
assertThat(response.getErrorCode()).isEqualTo(HandlerErrorCode.AlreadyExists);
```

[I've seen plenty of examples of the former](https://github.com/search?q=assertThrows+filename%3Acreatehandlertest.java+org%3Aaws-cloudformation&type=Code) but according to your suggestion we should use the latter?

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
miparnisari: > u can create a single call chain in base handler then use it in both handlers

https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-metricfilter/src/main/java/software/amazon/logs/metricfilter/BaseHandlerStd.java#L39-L57

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -29,157 +26,92 @@
...
             // STEP 1 [check if resource already exists]
             // if target API does not support 'ResourceAlreadyExistsException' then following check is required
             // for more information -> https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html
-            //.then(progress -> checkForPreCreateResourceExistence(proxy, request, progress))
+            .then(progress ->
+                // STEP 1.0 [initialize a proxy context]
+                // If your service API is not idempotent, meaning it does not distinguish duplicate create requests against some identifier (e.g; resource Name)
+                // and instead returns a 200 even though a resource already exists, you must first check if the resource exists here
+                // NOTE: If your service API throws 'ResourceAlreadyExistsException' for create requests this method is not necessary
+                proxy.initiate("{{ call_graph }}::{{ operation }}::PreExistanceCheck", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
```
miparnisari, line 42:
```suggestion
                proxy.initiate("{{ call_graph }}::{{ operation }}::PreExistenceCheck", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
```

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -29,157 +26,92 @@
...
+
+                    // STEP 1.2 [TODO: make an api call]
+                    .makeServiceCall((awsRequest, client) -> {
+                        AwsResponse awsResponse = null;
+
+                        // TODO: add custom read resource logic
+
+                        logger.log(String.format("%s has successfully been read.", ResourceModel.TYPE_NAME));
+                        return awsResponse;
+                    })
```
miparnisari, line 55:
```suggestion
                    .makeServiceCall((awsRequest, sdkProxyClient) -> sdkProxyClient.injectCredentialsAndInvokeV2(awsRequest, sdkProxyClient.client()::describeResource))
```

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -26,126 +25,88 @@
...
 
-        return ProgressEvent.progress(model, callbackContext)
+        return ProgressEvent.progress(request.getDesiredResourceState(), callbackContext)
             // STEP 1 [first update/stabilize progress chain - required for resource update]
             .then(progress ->
                 // STEP 1.0 [initialize a proxy context]
-                proxy.initiate("{{ call_graph }}::{{ operation }}", proxyClient, model, callbackContext)
+                // Implement client invocation of the update request through the proxyClient, which is already initialised with
+                // caller credentials, correct region and retry settings
+                proxy.initiate("{{ call_graph }}::{{ operation }}::first", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
```
miparnisari, line 37: ::first
```suggestion
                proxy.initiate("{{ call_graph }}::{{ operation }}::sendUpdateRequest", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
```

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -26,126 +25,88 @@
...
                     .progress())
 
             // If your resource is provisioned through multiple API calls, then the following pattern is required (and might take as many postUpdate callbacks as necessary)
             // STEP 2 [second update/stabilize progress chain]
             .then(progress ->
                     // STEP 2.0 [initialize a proxy context]
-                    proxy.initiate("{{ call_graph }}::{{ operation }}", proxyClient, model, callbackContext)
+                    // If your resource is provisioned through multiple API calls, you will need to apply each subsequent update
+                    // step in a discrete call/stabilize chain to ensure the entire resource is provisioned as intended.
+                    proxy.initiate("{{ call_graph }}::{{ operation }}::second", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
```
miparnisari, line 82:
```suggestion
                    proxy.initiate("{{ call_graph }}::{{ operation }}::stabilize", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
```

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
ammokhov: not sure this will build bc sdkProxyClient isnt defined in mocked example, so there is no `describeResource` method

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
ammokhov: I meant that it's the first update in case of multiple update calls required

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
ammokhov: same here, it's not really for stabilization. it's for the second update call as a rough example

python/rpdk/java/templates/init/guided_aws/StubCreateHandler.java
```diff
@@ -29,157 +26,92 @@
 
         this.logger = logger;
 
-        final ResourceModel model = request.getDesiredResourceState();
-
         // TODO: Adjust Progress Chain according to your implementation
         // https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/proxy/CallChain.java
 
-        return ProgressEvent.progress(model, callbackContext)
+        return ProgressEvent.progress(request.getDesiredResourceState(), callbackContext)
```
jotompki, line 32: now the model isnt getting set ? how is primary identifier going to get updated

PR #268, opened 2020-05-21
Cloudtrail log request id and request name.

goughcor:
*Description of changes:*
Log request name and request id.


rjlohan:
> I added it to injectCredentialsAndInvoke in the latest commit. What about the other methods? (Ex: injectCredentialsAndInvokeV2InputStream) Does it need to be added there too?

Yeah you'll probably want to wrap all the calls like `return requestFunction.apply(wrappedRequest);` to get full coverage from this logging.

goughcor:
> > I added it to injectCredentialsAndInvoke in the latest commit. What about the other methods? (Ex: injectCredentialsAndInvokeV2InputStream) Does it need to be added there too?
> 
> Yeah you'll probably want to wrap all the calls like `return requestFunction.apply(wrappedRequest);` to get full coverage from this logging.

I added logging to the other calls.  We should have full coverage now.

Issue #273, opened 2020-06-05
Validate resource before deserialize into resource model


wbingli:
As for now,  Java plugin will deserialize the raw resource property into resource model before validation. 

If the resource property is an invalid structure, e.g. pass a String but schema requires a List, it will fail sadly. And CloudFormation will show `Internal Failure` for this resource.

If validate resource before deserialize it, it can validate such invalid resource properly.

rjlohan:
Before making changes here, verify the history of the serialization behaviour. e.g;

* https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/commit/5e7fe4ea2163347923cbb9eff4c5d899e05fc050#diff-f71d0cfe4773c081286e31b592b040ff

* https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/commit/7f221a22f1e00a037ccd36f8e357e729c7880280#diff-f71d0cfe4773c081286e31b592b040ff

Validating the raw JSON can produce other errors, or fail with false negatives if it's not ordered correctly.

wbingli:
> Before making changes here, verify the history of the serialization behaviour. e.g;
> 
> * [5e7fe4e#diff-f71d0cfe4773c081286e31b592b040ff](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/commit/5e7fe4ea2163347923cbb9eff4c5d899e05fc050#diff-f71d0cfe4773c081286e31b592b040ff)
> * [7f221a2#diff-f71d0cfe4773c081286e31b592b040ff](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/commit/7f221a22f1e00a037ccd36f8e357e729c7880280#diff-f71d0cfe4773c081286e31b592b040ff)
> 
> Validating the raw JSON can produce other errors, or fail with false negatives if it's not ordered correctly.

The PR #274 do consider for such case.  It will only try to catch `MismatchedInputException` if the first deserialization fails due to mismatch type.   This avoid the false negative for the stringifier issue but provide useful information if type mismatch.

The following validation is kept as-is so that it will validate based on the converted resource model.

wbingli:
Resolving as the fix is merged.

PR #274, opened 2020-06-08
Validate resource model if type mismatch

wbingli:
*Issue #, if available:* #273 

*Description of changes:*

This targets to solve the problem #273, which the Java plugin will fail sadly if the resource model type mismatch.e.g. needs a String but giving a List.  Customer will see `Internal Failure` for such issue, it's pretty bad experience to debug such error, which needs cloudformation service support.

The whole fix is not trying to refactor the way we are doing validation now, but try to solve this specified problem.  Therefore, the change and risk is minimal but result is what we expected.

__Why change branch coverage to 0.80__
The build failed due to branch coverage. But this change covers the code change, it's due to the whole all branch increase but other classes has low branch coverage. 

### Testing

Tested with `LogGroup` with an invalid property:

The `RetentionInDays` should an number not a list.
```
  LogGroup:
    Type: 'AWS::Logs::LogGroup'
    Properties:
      RetentionInDays:
        - 1
```

Got expected validation message:

```

2020-06-08 01:37:42 | AWS::Logs::LogGroup | LogGroup |   | CREATE_FAILED | Model validation failed (#/RetentionInDays: #: only 0 subschema matches out of 2 #/RetentionInDays: expected type: Number, found: JSONArray #/RetentionInDays: failed validation constraint for keyword [enum]) |



```

### DIFF
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -20,6 +20,7 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.LambdaLogger;
 import com.amazonaws.services.lambda.runtime.RequestStreamHandler;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.exc.MismatchedInputException;
 import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
@@ -243,6 +244,7 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
         this.lambdaLogger = context.getLogger();
         ProgressEvent<ResourceT, CallbackT> handlerResponse = null;
         HandlerRequest<ResourceT, CallbackT> request = null;
+        String bearerToken = null;
         scrubFiles();
         try {
             if (inputStream == null) {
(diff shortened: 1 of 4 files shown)


src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -252,9 +254,16 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
...
             JSONObject rawInput = new JSONObject(new JSONTokener(input));
+            bearerToken = rawInput.getString("bearerToken");
 
             // deserialize incoming payload to modelled request
-            request = this.serializer.deserialize(input, typeReference);
+            try {
+                request = this.serializer.deserialize(input, typeReference);
+            } catch (MismatchedInputException e) {
+                JSONObject resourceSchemaJSONObject = provideResourceSchemaJSONObject();
+                JSONObject rawModelObject = rawInput.getJSONObject("requestData").getJSONObject("resourceProperties");
```
rjlohan, line 264: What happens to exceptions thrown here, if the JSON payload is completely malformed and doesn't contain the fields you're expecting?

wbingli: 1. The JSON payload completely malformed only if it invoked not by CloudFormation, which this case is really unlikely.

2. The above `this.serializer.deserialize(input, typeReference)` will throw different exception if the payload is malformed, it won't be catched by `MismatchedInputException`. We have unit test to cover the malformed input (0 bytes or invalid json),  those passed as expected.

3.  The only case I can think that `rawInput.getJSONObject("requestData").getJSONObject("resourceProperties");` is `null` or malformed `requestData`, the whole payload `HandlerRequest` has `MismatchedInputException`  in other fields.  However, I don't think it's case we need to worry about.

PR #275, opened 2020-06-10
Log error when failing to schedule re-invocation

arnulfojr:
*Description of changes:*

At the moment, when the handler fails to schedule a re-invocation the error message is not displayed in any manner and is hard to guess what was the error of the stack when the handlers succeed on "handling" the resource operation.

This at least will make it easier for the operator to know the details of the error.

> Not a required change, just making life easier in the meantime


rjlohan:
This change might be made redundant by https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/279 as we have changed the model for re-invocation to allow us to increase some limits (which may have been the very things causing the re-invoke to fail for you).

PR #278, opened 2020-06-12
added generic  type for multiple type support

ammokhov:
*Issue #, if available:*

*Description of changes:*

travis build fails because it requires this to get merged https://github.com/aws-cloudformation/cloudformation-cli/pull/454

### DIFF
--- a/python/rpdk/java/resolver.py
+++ b/python/rpdk/java/resolver.py
@@ -1,10 +1,11 @@
-from rpdk.core.jsonutils.resolver import UNDEFINED, ContainerType
+from rpdk.core.jsonutils.resolver import MULTIPLE, UNDEFINED, ContainerType
 
 PRIMITIVE_TYPES = {
     "string": "String",
     "integer": "Integer",
     "boolean": "Boolean",
     "number": "Double",
+    MULTIPLE: "Object",
     UNDEFINED: "Object",
 }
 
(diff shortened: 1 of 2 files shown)


python/rpdk/java/resolver.py
```diff
@@ -1,10 +1,11 @@
-from rpdk.core.jsonutils.resolver import UNDEFINED, ContainerType
+from rpdk.core.jsonutils.resolver import MULTIPLE, UNDEFINED, ContainerType
 
 PRIMITIVE_TYPES = {
     "string": "String",
     "integer": "Integer",
     "boolean": "Boolean",
     "number": "Double",
+    MULTIPLE: "Object",
```
wbingli, line 8: If we can define a new resolved_type container type, it can be like:

```
    if resolved_type.container == ContainerType.MULTIPLE:
        return "Object"
```

python/rpdk/java/resolver.py
```diff
@@ -8,13 +8,20 @@
...
+
 
 def translate_type(resolved_type):
     if resolved_type.container == ContainerType.MODEL:
         return resolved_type.type
     if resolved_type.container == ContainerType.PRIMITIVE:
         return PRIMITIVE_TYPES[resolved_type.type]
 
+    if resolved_type.container == ContainerType.MULTIPLE:
+        return MULTIPLE_TYPES[resolved_type.type]
```
wbingli, line 23: No need to have `MULTIPE_TYPES`. It can be just `return "Object"`.

PR #283, opened 2020-07-07
Fix for removing delay and handshake mode from the codebase

dchakrav-github:
Removing 60s wait time from the chain for handshake mode.

### DIFF
--- a/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
+++ b/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
@@ -40,6 +40,8 @@ import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
 import software.amazon.awssdk.awscore.AwsResponse;
 import software.amazon.awssdk.awscore.exception.AwsErrorDetails;
 import software.amazon.awssdk.awscore.exception.AwsServiceException;
+import software.amazon.awssdk.core.ResponseBytes;
+import software.amazon.awssdk.core.ResponseInputStream;
 import software.amazon.awssdk.core.SdkClient;
 import software.amazon.awssdk.core.exception.NonRetryableException;
 import software.amazon.awssdk.core.exception.RetryableException;
@@ -122,6 +124,22 @@ public class AmazonWebServicesClientProxy implements CallChain {
                 IterableT
                 injectCredentialsAndInvokeIterableV2(RequestT request, Function<RequestT, IterableT> requestFunction) {
                 return AmazonWebServicesClientProxy.this.injectCredentialsAndInvokeIterableV2(request, requestFunction);
+            }
+
+            @Override
+            public <RequestT extends AwsRequest, ResponseT extends AwsResponse>
+                ResponseInputStream<ResponseT>
+                injectCredentialsAndInvokeV2InputStream(RequestT request,
+                                                        Function<RequestT, ResponseInputStream<ResponseT>> requestFunction) {
+                return AmazonWebServicesClientProxy.this.injectCredentialsAndInvokeV2InputStream(request, requestFunction);
+            }
+
+            @Override
+            public <RequestT extends AwsRequest, ResponseT extends AwsResponse>
+                ResponseBytes<ResponseT>
+                injectCredentialsAndInvokeV2Bytes(RequestT request,
+                                                  Function<RequestT, ResponseBytes<ResponseT>> requestFunction) {
+                return AmazonWebServicesClientProxy.this.injectCredentialsAndInvokeV2Bytes(request, requestFunction);
             }
 
             @Override
@@ -448,7 +466,9 @@ public class AmazonWebServicesClientProxy implements CallChain {
         request.setRequestCredentialsProvider(v1CredentialsProvider);
 
         try {
-            return requestFunction.apply(request);
+            ResultT respose = requestFunction.apply(request);
+            logRequestMetadata(request, respose);
+            return respose;
         } catch (final Throwable e) {
             loggerProxy.log(String.format("Failed to execute remote function: {%s}", e.getMessage()));
             throw e;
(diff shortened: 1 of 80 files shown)


dchakrav-github:
> LGTM.
> 
> Same question as in the email discussion, I think we should remove the local re-try if it's only for re-try the stabilization call(https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java#L425-L431), we are leveraging the service to make re-invoke.

We should maximize the time we have for the invoke when possible. If we have sufficent time to locally wait and re-invoke then we should use that time as the context is there to help. The loops can get expensvie if we had a simple flat model of 2s wait and retry for 60s, it would be expensive to roundtrip all of those calls. This also works consistently across any compute model we might bind to and later transparently send out progress event, e.g. HTTP/2 push model on containers.

wbingli:
> We should maximize the time we have for the invoke when possible. If we have sufficent time to locally wait and re-invoke then we should use that time as the context is there to help.

As for now, we should be careful to maximize the time inside the lambda function. As it's have 3 mins for timeout but invocation timeout is 70 seconds.  That means the local retry can be still running when backend consider it's timeout.

> The loops can get expensvie if we had a simple flat model of 2s wait and retry for 60s, it would be expensive to roundtrip all of those calls. 

It's arguable to say the roundtrip is more expensive.  The lambda function is called synchronously, which will block backend resources (threads).   The time it sleeps in the lambda function, it blocks the backend service resource to be used for handling other requests, which reduces the whole service throughput.  The backend re-invocation is event based, in current low latency network especially inside same region, it's mostly single digit milliseconds for a call, the network roundtrip would be small frication comparing to even 2s wait time.

>  This also works consistently across any compute model we might bind to and later transparently send out progress event, e.g. HTTP/2 push model on containers.

That would be a different story if we need to introduce a different communication model.  Speaking to the computing model,   current implementation depends on lambda to provide remaining time information.  As for other computing model, I don't know if we should have such assumption that they can provide remaining time information inside the execution time.

dchakrav-github:
> > We should maximize the time we have for the invoke when possible. If we have sufficent time to locally wait and re-invoke then we should use that time as the context is there to help.
> 
> As for now, we should be careful to maximize the time inside the lambda function. As it's have 3 mins for timeout but invocation timeout is 70 seconds. That means the local retry can be still running when backend consider it's timeout.

SLG = Service Level Grant (this is the grant the client provides for the API call, that must be >= MSLA)
MSLA = Minimum Service Level Agreement (this is the mimum the client must specify for an API call. This is the bare minimum the server needs to make some progress for the API call)

Why are the values for time to execute and the promised SLG for the invocation different? If we make the time to run <= client timeout, that just makes it consistent with how any API call behaves. The client provides the SLG that must be met for the API invocation. The server tries to meet the SLG. If it can't it will respond with an IN_PROGRESS callback to complete. It is a very simple request/response system that works. There is minimum SLA (server provided) which the client can't violate for the API

> 
> > The loops can get expensvie if we had a simple flat model of 2s wait and retry for 60s, it would be expensive to roundtrip all of those calls.
> 
> It's arguable to say the roundtrip is more expensive. The lambda function is called synchronously, which will block backend resources (threads). The time it sleeps in the lambda function, it blocks the backend service resource to be used for handling other requests, which reduces the whole service throughput. The backend re-invocation is event based, in current low latency network especially inside same region, it's mostly single digit milliseconds for a call, the network roundtrip would be small frication comparing to even 2s wait time.
> 

There are several assumptions baked into this including Lambda cold starts and more for each invocation. The meta point I am making is there is set SLG for the API call the client is expecting the response to be satisfied within. Why is the client concerned with how the API operates beyond satisfying the given SLG? It is an implementation detail on how the client is making those calls and awaiting responses, which can be non-blocking event driven or threaded model. That does not affect how the server chooses to respond to the API call being made.

> > This also works consistently across any compute model we might bind to and later transparently send out progress event, e.g. HTTP/2 push model on containers.
> 
> That would be a different story if we need to introduce a different communication model. Speaking to the computing model, current implementation depends on lambda to provide remaining time information. As for other computing model, I don't know if we should have such assumption that they can provide remaining time information inside the execution time.

It isn't a different story, the model is very similar to request/reply model. Here for the API call that is being made, the client still provides a SLG. The primary difference is the client does not wait for the reponse, it effectively receives push message responses as things proceed from the server, with a final push message with SUCCESS, FAIL or IN_PROGRESS state for the API call completion with the SLG provided. The client can then proceed as needed.

wbingli:
I agree that the lambda execution timeout should be less than client invocation timeout.  However, this will take time to be properly fixed.    Before that,  we cannot maximize the time in the lambda. 

> Why is the client concerned with how the API operates beyond satisfying the given SLG? It is an implementation detail on how the client is making those calls and awaiting responses, which can be non-blocking event driven or threaded model. That does not affect how the server chooses to respond to the API call being made.

Well, if thinking about optimization for the whole service,  we should consider the cost from the CFN service as client to invoke the handlers.   The blocking model reduces the service throughput and could impact the service availability and I don't think we should encourage such maximize to the limit of SLG.

Let's discuss offline to follow up.

dchakrav-github:
Made changes based on the discussion. Introduced a WaitStrategy that defaults to schedule callbacks back to service. An alternative local loop strategy is also tested for and used for other compute bindings

src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
```diff
@@ -422,15 +408,10 @@ public Logger getLogger() {
                                         return ProgressEvent.failed(model, context, HandlerErrorCode.NotStabilized,
                                             "Exceeded attempts to wait");
                                     }
-                                    long remainingTime = getRemainingTimeInMillis();
-                                    long localWait = next.toMillis() + 2 * elapsed + 100;
-                                    if (remainingTime > localWait) {
-                                        loggerProxy.log("Waiting for " + next.getSeconds() + " for call " + callGraph);
-                                        Uninterruptibles.sleepUninterruptibly(next.getSeconds(), TimeUnit.SECONDS);
-                                        continue;
+                                    event = AmazonWebServicesClientProxy.this.waitStrategy.await(elapsed, next, context, model);
```
wbingli, line 411: nit:  Would be good to use a different variable as  it's about whether to continue.  

```
final Optional<ProgressEvent> completeLocalExecution = AmazonWebServicesClientProxy.this.waitStrategy.await(elapsed, next, context, model);
if (completeLocalExecution.present()) {
 return completeLocalExecution.get();
}
```

Issue #209, opened 2019-11-20
Local lambda logging during contract tests all on one line


jotompki:
There's no line break after every log statement when running the sam cli and contract tests. End up with a dense one line log like this:
```
AWS::SES::ConfigurationSet [{"Name":["-"]}] deleted successfully[DELETE] handler invokedEND RequestId: dd5fb042-f25c-1a50-dd12-d410fa7311f8
```

Need to see what we can do to separate these out into individual lines. Should be a quick change.

PR #290, opened 2020-07-16
Adding newlines to logger

rjlohan:
*Issue #, if available:* #209 

*Description of changes:* Adding newline to LambdaLogPublisher so that logs are intelligible in contract test output in SAM Local

### DIFF
--- a/src/main/java/software/amazon/cloudformation/loggers/LambdaLogPublisher.java
+++ b/src/main/java/software/amazon/cloudformation/loggers/LambdaLogPublisher.java
@@ -28,6 +28,6 @@ public class LambdaLogPublisher extends LogPublisher {
 
     @Override
     protected void publishMessage(String message) {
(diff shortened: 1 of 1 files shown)


src/main/java/software/amazon/cloudformation/loggers/LambdaLogPublisher.java
```diff
@@ -28,6 +28,6 @@ public LambdaLogPublisher(final LambdaLogger logger,
 
     @Override
     protected void publishMessage(String message) {
-        this.logger.log(message);
+        this.logger.log(String.format("%s%n", message));
```
ammokhov, line 31: `%n` or `\n`?

rjlohan: `pre-commit` rules want it this way.

PR #291, opened 2020-07-24
Use standard AWS SDK exception message

wbingli:
*Issue #, if available:*

*Description of changes:*

The current error message is built with Java class name, which is ideal comparing with standard SDK exception message. It also miss important information such as request ID etc.

Just use standard SDK exception message which includes important information as well as not expose internal java class as event message.

__Event Message Before:__

```
Exception=[class software.amazon.awssdk.services.kms.model.KmsException] ErrorCode=[AccessDeniedException],  ErrorMessage=[User: arn:aws:sts::1234567890:assumed-role/abcd is not authorized to perform: kms:UntagResource on resource: arn:aws:kms:eu-central-1:1234567890:key/10b977e9-da47-4c82-929f-c3c3506709d0]
```

__Event Message After:__

```
User: arn:aws:sts::1234567890:assumed-role/abcd is not authorized to perform: kms:UntagResource on resource: arn:aws:kms:eu-central-1:1234567890:key/10b977e9-da47-4c82-929f-c3c3506709d0 (Service: AWSKMS; Status Code: 400;  Request ID: c86b48d0-2820-4ff3-a1de-d52ed24f922c)
```

### DIFF
--- a/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
+++ b/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
@@ -572,8 +572,7 @@ public class AmazonWebServicesClientProxy implements CallChain {
         if (e instanceof AwsServiceException) {
             AwsServiceException sdkException = (AwsServiceException) e;
             AwsErrorDetails details = sdkException.awsErrorDetails();
-            String errMsg = "Exception=[" + sdkException.getClass() + "] " + "ErrorCode=[" + details.errorCode()
-                + "],  ErrorMessage=[" + details.errorMessage() + "]";
+            String errMsg = sdkException.getMessage();
             switch (details.sdkHttpResponse().statusCode()) {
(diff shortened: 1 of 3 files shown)


wbingli:
> I'm actually not too concerned here. This 'internal implementation' is the AWS SDK, which is open source, being called from a Resource Type handler, which is also open source. There is no real meaningful information leak here, and we lose the structure in diagnostics with this change.

However, this structure is not standard SDK exception message, and it introduces a new structure comparing with all existing resources. It causes inconsistent for customer to read the error message between different resources.

In the meanwhile, exposing a full Java class exception name doesn't seems proper as service error. e.g. `Exception=[class software.amazon.awssdk.services.kms.model.KmsException]`. How much value for customer can get to diagnose their stack error? Especially the full class name,  it's clearly internal resource implementation, even the class is open source but we should not expose it.

Not all resource are public before they are published. This at least some level of information leaking to expose such low level error message to customer.

PR #293, opened 2020-07-29
Removed account ID from metric namespace

cortez7:
The removed account ID is the customer account ID instead of the provider account ID. This means that for public types, a new metric namespace would be created each time we need to emit metrics. The main issue with this is that the internal and logging failure alarms will not work properly and service teams may not be notified of issues.

Go plugin PR: https://github.com/aws-cloudformation/cloudformation-cli-go-plugin/pull/151
Python plugin PR: https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/pull/108


cortez7:
When I ran pre-commit run --all-files everything passed, although beforehand I hand to run `pip install pylint` and `pip install pytest-cov`. We decided to fix the build issue in a different PR. We may need to lock the versions.

cortez7:
Build failures are now fixed with new pytest release.

PR #294, opened 2020-08-01
added some pre-actionable checks for update and read handler to align…

ammokhov:
… with handler contract

*Issue #, if available:*

*Description of changes:*

Codegen enhancements to provide to a developer additional tools for contract test passing:

- [x] *_An update handler MUST return FAILED with a NotFound error code if the resource did not exist prior to the update request._*

- [x] *When the delete handler returns SUCCESS, the ProgressEvent object MUST NOT contain a model.*

- [x] *A read handler MUST return FAILED with a NotFound error code if the resource does not exist.*

### DIFF
--- a/python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
+++ b/python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
@@ -56,7 +56,7 @@ public class DeleteHandler extends BaseHandlerStd {
                     // STEP 1.3 [TODO: handle exception]
                     .handleError((awsRequest, exception, client, model, context) -> {
                         // TODO: uncomment when ready to implement
-                        // if (exception instanceof CfnNotFoundException)
+                        // if (exception instanceof ResourceNotFoundException)
                         //     return ProgressEvent.success(model, context);
                         // throw exception;
                         return ProgressEvent.progress(model, context);
@@ -105,6 +105,10 @@ public class DeleteHandler extends BaseHandlerStd {
                         logger.log(String.format("%s [%s] deletion has stabilized: %s", ResourceModel.TYPE_NAME, model.getPrimaryIdentifier(), stabilized));
                         return stabilized;
                     })
-                    .success());
+                    .progress()
+            )
+
+            // STEP 3 [TODO: return the successful progress event without resource model]
+            .then(progress -> ProgressEvent.defaultSuccessHandler(null));
     }
 }
--- a/python/rpdk/java/templates/init/guided_aws/StubHandlerTest.java
+++ b/python/rpdk/java/templates/init/guided_aws/StubHandlerTest.java
@@ -60,7 +60,11 @@ public class {{ operation }}HandlerTest extends AbstractTestBase {
         assertThat(response).isNotNull();
         assertThat(response.getStatus()).isEqualTo(OperationStatus.SUCCESS);
         assertThat(response.getCallbackDelaySeconds()).isEqualTo(0);
+        {% if operation == "Delete" %}
+        assertThat(response.getResourceModel()).isNull();
+        {% else %}
         assertThat(response.getResourceModel()).isEqualTo(request.getDesiredResourceState());
+        {% endif %}
         assertThat(response.getResourceModels()).isNull();
         assertThat(response.getMessage()).isNull();
         assertThat(response.getErrorCode()).isNull();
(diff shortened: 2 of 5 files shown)


python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
```diff
@@ -105,6 +105,10 @@
                         logger.log(String.format("%s [%s] deletion has stabilized: %s", ResourceModel.TYPE_NAME, model.getPrimaryIdentifier(), stabilized));
                         return stabilized;
                     })
-                    .success());
+                    .progress()
+            )
+
+            // STEP 3 [TODO: return the successful progress event without resource model]
+            .then(progress -> ProgressEvent.defaultSuccessHandler(null));
```
anshikg, line 112: lets return a progress event without the model.

python/rpdk/java/templates/init/guided_aws/StubHandlerTest.java
```diff
@@ -60,7 +60,11 @@ public void handleRequest_SimpleSuccess() {
         assertThat(response).isNotNull();
         assertThat(response.getStatus()).isEqualTo(OperationStatus.SUCCESS);
         assertThat(response.getCallbackDelaySeconds()).isEqualTo(0);
+        {% if operation == "Delete" %}
+        assertThat(response.getResourceModel()).isNull();
+        {% else %}
```
anshikg, line 65: 👍

python/rpdk/java/templates/init/guided_aws/Translator.java
```diff
@@ -52,6 +52,7 @@ static ResourceModel translateFromReadResponse(final AwsResponse awsResponse) {
     // e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/2077c92299aeb9a68ae8f4418b5e932b12a8b186/aws-logs-loggroup/src/main/java/com/aws/logs/loggroup/Translator.java#L58-L73
     return ResourceModel.builder()
         //.someProperty(response.property())
+        // IMPORTANT: exclude writeOnlyProperties
```
anshikg, line 55: isn't this done by lambda wrapper?

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -29,17 +29,45 @@
         // https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/proxy/CallChain.java
 
         return ProgressEvent.progress(request.getDesiredResourceState(), callbackContext)
-            // STEP 1 [first update/stabilize progress chain - required for resource update]
+
+            // STEP 1 [check if resource already exists]
+            // for more information -> https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html
+            // if target API does not support 'ResourceNotFoundException' then following check is required
```
anshikg, line 35: it seems we are checking for not found at multiple places. May be consolidate this to be in on place.

python/rpdk/java/templates/init/guided_aws/StubDeleteHandler.java
ammokhov: `ProgressEvent.defaultSuccessHandler(null)` returns without a model

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
ammokhov: we might be having same callGraph value cached in case you cross handler interaction. also I think it's already quite wordy if we refer in multiple places it will be confusing

python/rpdk/java/templates/init/guided_aws/Translator.java
rjlohan: Yes it should be. At least in this language plugin.

anshikg: could you just remove this line about write only properties. This might cause some confusion.

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -29,17 +29,45 @@
...
             .then(progress ->
                 // STEP 1.0 [initialize a proxy context]
+                // If your service API does not return ResourceNotFoundException on update requests against some identifier (e.g; resource Name)
+                // and instead returns a 200 even though a resource does not exist, you must first check if the resource exists here
+                // NOTE: If your service API throws 'ResourceNotFoundException' for update requests this method is not necessary
```
PatMyron, line 40: Even if the service API throws `ResourceNotFoundException`, the contract tests as currently implemented expect `ResourceNotFoundException` to be thrown even when calling the handlers with obviously invalid inputs which likely will hit model validation before checking if the resource exists or not, even for services that do throw `ResourceNotFoundException`

also discussed in https://github.com/aws-cloudformation/cloudformation-cli/issues/527

PR #296, opened 2020-08-04
Added writeOnlyProperties redaction to testEntrypoint

rjlohan:
This fix allows the contract tests to correctly assert on writeOnlyProperties being removed from the model.

*Issue #, if available:* N/A

*Description of changes:* Contract tests were reporting false negatives for the redaction of `writeOnlyProperties`. Minor refactoring so that the `testEntrypoint` also sanitizes the model by default.


rjlohan:
> This all looks to be good, but I think we should look into just removing the test entrypoint altogether since we have a synchronous response protocol now with the v2 protocol version. Would ease the pain of maintaining these two entrypoints.

Yeah that makes sense.

Issue #297, opened 2020-08-13
Publish by-HandlerErrorCode, by-OperationStatus and top-level InternalFailure metrics


clemengi:
Hi,

We would like to request to publish by-ErrorCode and by-OperationStatus metrics to allow the configuration of Cloudwatch alarms.

**Problem**

The current LambdaWrapper implementation publishes two sets of metrics related to anomalies:
```
  Namespace: AWS/CloudFormation/<FullyQualifiedResourceNameWithSlashes>
  Metric name: "HandlerException"
  Dimensions: Action, ExceptionType, HandlerErrorCode, ResourceType
```

```
  Namespace: AWS/CloudFormation/<AWS_ACCOUNT_ID>/<FullyQualifiedResourceNameWithSlashes>
  Metric name: HandlerException
  Dimensions: Action, ExceptionType, HandlerErrorCode, ResourceType
```

These metrics are great for providing a quick insight into a problem but they are not a good option for alarms, because the values of the ExceptionType dimension are not known in advance.

We would like to have additional metrics of the form:
```
Namespace:  <does not really matter>
Metric name: "HandlerException"
Dimensions: Action, HandlerErrorCode, ResourceType # with HandlerErrorCode in https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract-errors.html

Namespace:  <does not really matter>
Metric name: "HandlerException"
Dimensions: Action, OperationStatus, ResourceType # with OperationStatus in software.amazon.cloudformation.proxy.OperationStatus.values()

Namespace:  <does not really matter>
Metric name: "HandlerException"
Dimensions: Action, ResourceType
```

The first set of metrics would allow to set up alarms on a closed set of metrics, broken down by the HandlerErrorCode taxonomy.

The second metric would be useful for broad alarms on Progress status (e.g. spikes in FAILED or IN_PROGRESS calls).

The third metric should assume value 1 anytime a call to the Resource's Handler throws an exception, and zero otherwise. It would allow to alarm on any uncaught Exception, regardless of the type.

Thank you!

PR #298, opened 2020-08-18
additional cw-metrics

ammokhov:
*Issue #, if available:* #297 

*Description of changes:*

Added some extra metrics so to create cw alarms upon. 

**Metric No. 1**
```
Namespace: under resource type namespace
Metric name: "HandlerException"
Dimensions: Action, HandlerErrorCode, ResourceType 
```

**Metric No. 2** - not added: the reason being is that operation status for uncaught exceptions is always `FAILED`, so this metrics seems to be redundant; we have enough of the exact same information in other dimensions;
```
Namespace:  <does not really matter>
Metric name: "HandlerException"
Dimensions: Action, OperationStatus, ResourceType # with OperationStatus in software.amazon.cloudformation.proxy.OperationStatus.values()
```

**Metric No. 3**
```
Namespace:  under resource type namespace
Metric name: "HandlerException"
Dimensions: Action, ResourceType
```

### DIFF
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -312,6 +312,8 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
 
         if (handlerResponse.getStatus() == OperationStatus.IN_PROGRESS && !isMutatingAction) {
             throw new TerminalException("READ and LIST handlers must return synchronously.");
+        } else if (handlerResponse.getStatus() != OperationStatus.FAILED) {
+            this.metricsPublisherProxy.publishExceptionCountMetric(Instant.now(), request.getAction(), false);
         }
 
         return handlerResponse;
@@ -344,6 +346,11 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
             } else {
                 this.log("Handler returned null");
                 throw new TerminalException("Handler failed to provide a response.");
+            }
+
+            if (handlerResponse.getStatus() == OperationStatus.FAILED) {
+                publishExceptionMetric(request.getAction(), new Throwable(handlerResponse.getMessage()),
+                    handlerResponse.getErrorCode());
             }
 
             return handlerResponse;
@@ -464,6 +471,9 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
     private void publishExceptionMetric(final Action action, final Throwable ex, final HandlerErrorCode handlerErrorCode) {
         if (this.metricsPublisherProxy != null) {
             this.metricsPublisherProxy.publishExceptionMetric(Instant.now(), action, ex, handlerErrorCode);
+            this.metricsPublisherProxy.publishExceptionByErrorCodeMetric(Instant.now(), action, handlerErrorCode);
+            // always puts failed events
+            this.metricsPublisherProxy.publishExceptionCountMetric(Instant.now(), action, true);
         } else {
             // Lambda logger is the only fallback if metrics publisher proxy is not
             // initialized.
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -29,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -313,7 +314,7 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
         if (handlerResponse.getStatus() == OperationStatus.IN_PROGRESS && !isMutatingAction) {
             throw new TerminalException("READ and LIST handlers must return synchronously.");
         } else if (handlerResponse.getStatus() != OperationStatus.FAILED) {
-            this.metricsPublisherProxy.publishExceptionCountMetric(Instant.now(), request.getAction(), false);
+            publishExceptionCodeAndCountMetric(request.getAction(), handlerResponse.getErrorCode(), false);
         }
 
         return handlerResponse;
@@ -349,8 +350,7 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
             }
 
             if (handlerResponse.getStatus() == OperationStatus.FAILED) {
-                publishExceptionMetric(request.getAction(), new Throwable(handlerResponse.getMessage()),
-                    handlerResponse.getErrorCode());
+                publishExceptionCodeAndCountMetric(request.getAction(), handlerResponse.getErrorCode(), true);
             }
 
             return handlerResponse;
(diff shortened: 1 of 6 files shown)


src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -464,6 +471,9 @@ protected void validateModel(final JSONObject modelObject) throws ValidationExce
     private void publishExceptionMetric(final Action action, final Throwable ex, final HandlerErrorCode handlerErrorCode) {
         if (this.metricsPublisherProxy != null) {
             this.metricsPublisherProxy.publishExceptionMetric(Instant.now(), action, ex, handlerErrorCode);
+            this.metricsPublisherProxy.publishExceptionByErrorCodeMetric(Instant.now(), action, handlerErrorCode);
+            // always puts failed events
+            this.metricsPublisherProxy.publishExceptionCountMetric(Instant.now(), action, true);
```
clemengi, line 476: I guess this would require a larger refactor, but ideally a 0/1 metric should be emitted in a single place, because we want to be sure that sum(data points with value = 0) + sum(data point with value = 1) = (number of calls to monitored method). (because then the average of that metric represents the error rate)

Emitting false in one place and true in another makes hard to guarantee that, for a single call to the monitored method, exactly one data point is emitted.

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -346,6 +348,11 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
                 throw new TerminalException("Handler failed to provide a response.");
             }
 
+            if (handlerResponse.getStatus() == OperationStatus.FAILED) {
+                publishExceptionMetric(request.getAction(), new Throwable(handlerResponse.getMessage()),
```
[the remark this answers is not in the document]
ammokhov: yea, but message is not really a part of the requested metrices bc it's not static and varies upon multiple factors; the real values used in this case are `HandlerErrorCode` and `Action`

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
ammokhov: this is the method for any exception thrown, so it handles any failures as `1`; otherwise when response is successful and not equal to `FAILED` status it emits `0`

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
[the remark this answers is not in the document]
ammokhov: I think `null` will be more appropriate. will change that

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -346,6 +349,10 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
                 throw new TerminalException("Handler failed to provide a response.");
             }
 
+            if (handlerResponse.getStatus() == OperationStatus.FAILED) {
+                publishExceptionCodeAndCountMetric(request.getAction(), handlerResponse.getErrorCode(), true);
```
wbingli, line 353: Here the problem is that if the Operation fails,  only the error code has metric data point as 1. But we are missing the sample data for other error code as 0. Which means, The totally data points is not right.

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -464,13 +471,27 @@ protected void validateModel(final JSONObject modelObject) throws ValidationExce
     private void publishExceptionMetric(final Action action, final Throwable ex, final HandlerErrorCode handlerErrorCode) {
         if (this.metricsPublisherProxy != null) {
             this.metricsPublisherProxy.publishExceptionMetric(Instant.now(), action, ex, handlerErrorCode);
+            // always puts failed events
+            publishExceptionCodeAndCountMetric(action, handlerErrorCode, true);
```
wbingli, line 475: We have 3 places to publish the metrics for the error code.  I think we should merge them into one, just to have it after `wrapInvocationAndHandleErrors`.

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -312,6 +313,8 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
 
         if (handlerResponse.getStatus() == OperationStatus.IN_PROGRESS && !isMutatingAction) {
             throw new TerminalException("READ and LIST handlers must return synchronously.");
+        } else if (handlerResponse.getStatus() != OperationStatus.FAILED) {
+            publishExceptionCodeAndCountMetric(request.getAction(), handlerResponse.getErrorCode(), false);
```
wbingli, line 317: To simplify the whole metric publish,   how about below:
```


EnumSet.allOf(HandlerErrorCode.class).stream()
            .filter(errorCode -> errorCode.equals(handlerErrorCode)).forEach(errorCode -> this.metricsPublisherProxy
                .publishExceptionByErrorCodeMetric(Instant.now(), action, errorCode, errorCode.equals(handlerResponse.getErrorCode)));
        this.metricsPublisherProxy.publishExceptionCountMetric(Instant.now(), action, handlerResponse.getStatus() != OperationStatus.FAILED);
```

src/test/java/software/amazon/cloudformation/LambdaWrapperTest.java
```diff
@@ -399,14 +402,21 @@ public void invokeHandler_InProgress_returnsInProgress(final String requestDataP
                 // verify output response
                 verifyHandlerResponse(out, ProgressEvent.<TestModel, TestContext>builder().status(OperationStatus.IN_PROGRESS)
                     .resourceModel(TestModel.builder().property1("abc").property2(123).build()).build());
+                verify(providerMetricsPublisher, atLeastOnce()).publishExceptionByErrorCodeMetric(any(Instant.class), eq(action),
```
wbingli, line 405: As we have the total known ErrorCode, we should know exactly how many times

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -471,6 +477,21 @@ private void publishExceptionMetric(final Action action, final Throwable ex, fin
...
 
+    /*
+     * null-safe exception metrics delivery
+     */
+    private void
+        publishExceptionCodeAndCountMetric(final Action action, final HandlerErrorCode handlerErrorCode, final boolean thrown) {
+        if (this.metricsPublisherProxy != null) {
+            EnumSet.allOf(HandlerErrorCode.class).stream()
+                // publishing 0 value for all (if not thrown) otherwise filtered
+                .filter(errorCode -> errorCode.equals(handlerErrorCode) || !thrown)
```
wbingli, line 488: This line can be removed.  We actually want to do forEach for all ErrorCode, not match in what cases.

In case of in a Failed cases, this will only emit metric 1 for the one equals to the ErrorCode, but won't emit 0 metric for other metrics.

src/main/java/software/amazon/cloudformation/metrics/MetricsPublisherImpl.java
```diff
@@ -66,6 +66,27 @@ public void publishExceptionMetric(final Instant timestamp,
...
     }
 
+    @Override
+    public void publishExceptionByErrorCodeMetric(final Instant timestamp,
+                                                  final Action action,
+                                                  final HandlerErrorCode handlerErrorCode,
+                                                  final boolean thrown) {
+        Map<String, String> dimensions = new HashMap<>();
+        dimensions.put(Metric.DIMENSION_KEY_ACTION_TYPE, action == null ? "NO_ACTION" : action.name());
+        dimensions.put(Metric.DIMENSION_KEY_HANDLER_ERROR_CODE, handlerErrorCode.name());
```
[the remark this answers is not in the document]
wbingli: all right, the namespace actually has the resource type info already.

Issue #292, opened 2020-07-29
Provide `getPreviousResourceTags` in ResourceHandlerRequest


wbingli:
For update,  we should provide `getPreviousResourceTags` which comes from previousStackTags + previousResourceDefinedTags.

ammokhov:
closing as fixed by #300

PR #300, opened 2020-08-19
Add getPreviousResourceTags method 

omkhegde:
Issue #, if available: #292
    
Description of changes: This change adds getPreviousResourceTags method which returns previous tags that were applied to a resource. The tags come from previousStackTags and previousResourceDefinedTags

### DIFF
--- a/python/rpdk/java/templates/generate/HandlerWrapper.java
+++ b/python/rpdk/java/templates/generate/HandlerWrapper.java
@@ -132,7 +132,6 @@ public final class HandlerWrapper extends LambdaWrapper<{{ pojo_name }}, Callbac
             .desiredResourceState(requestData.getResourceProperties())
             .previousResourceState(requestData.getPreviousResourceProperties())
             .desiredResourceTags(getDesiredResourceTags(request))
-            .previousResourceTags(getPreviousResourceTags(request))
             .systemTags(request.getRequestData().getSystemTags())
             .awsAccountId(request.getAwsAccountId())
             .logicalResourceIdentifier(request.getRequestData().getLogicalResourceId())
--- a/python/rpdk/java/templates/generate/HandlerWrapper.java
+++ b/python/rpdk/java/templates/generate/HandlerWrapper.java
@@ -132,6 +132,7 @@ public final class HandlerWrapper extends LambdaWrapper<{{ pojo_name }}, Callbac
             .desiredResourceState(requestData.getResourceProperties())
             .previousResourceState(requestData.getPreviousResourceProperties())
             .desiredResourceTags(getDesiredResourceTags(request))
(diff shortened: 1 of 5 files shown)


wbingli:
LGTM.   Let's verify this change with one of the resources e.g. R53 HostedZone. Once we confirm it works (Update handler can get the previousResourceTags), will approve it.

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -521,6 +521,29 @@ protected void scrubFiles() {
...
+     *            tags which are specified as a direct resource property.
+     * @return a Map of Tag names to Tag values
+     */
+    @VisibleForTesting
+    protected Map<String, String> getPreviousResourceTags(final HandlerRequest<ResourceT, CallbackT> request) {
+        Map<String, String> previousResourceTags = new HashMap<>();
+
+        if (request != null && request.getRequestData() != null) {
+            replaceInMap(previousResourceTags, request.getRequestData().getPreviousStackTags());
+            replaceInMap(previousResourceTags,
```
wbingli, line 540: The `getPreviousResourceProperties()` is null for CREATE/DELETE/READ,  and not all `provideResourceDefinedTags` handles the null resource model. let's have a null check here.

```
if (request.getRequestData().getPreviousResourceProperties() != null) {
 replaceInMap(previousResourceTags,provideResourceDefinedTags(request.getRequestData().getPreviousResourceProperties()));
}

```

python/rpdk/java/templates/generate/HandlerWrapper.java
```diff
@@ -132,6 +132,7 @@ public JSONObject provideResourceSchemaJSONObject() {
             .desiredResourceState(requestData.getResourceProperties())
             .previousResourceState(requestData.getPreviousResourceProperties())
             .desiredResourceTags(getDesiredResourceTags(request))
+            .previousResourceTags(getPreviousResourceTags(request))
```
wbingli, line 135: This method is a new method. That mean anyone who updates the PyPi java plugin but without update the latest maven java plugin will run into problem, and it will break the build.

I'm afraid we need to copy the entire method `protected Map<String, String> getPreviousResourceTags(final HandlerRequest<ResourceT, CallbackT> request) {` into this HandlerWrapper as well as `replaceInMap`, which is a private method. 

It will be a little bit ugly. But comparing with breaking customers and wasting time to find and upgrade java plugin, I more prefer this temporary duplication.  

The entire `transform` method in the generated code seems not a good idea at all, but let's leave for further refactoring.

wbingli: Hmm, never mind, even we can duplicate this new method, but `getPreviousStackTags` still a new method and requires the Java plugin support.

PR #301, opened 2020-08-21
added zero-events on untriggered codes

ammokhov:
*Issue #, if available:*

*Description of changes:*

a follow up for the previous pr #298 => adding 0 values for the codes that did not cause exception

### DIFF
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -487,11 +487,8 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
     private void
         publishExceptionCodeAndCountMetric(final Action action, final HandlerErrorCode handlerErrorCode, final boolean thrown) {
         if (this.metricsPublisherProxy != null) {
-            EnumSet.allOf(HandlerErrorCode.class).stream()
-                // publishing 0 value for all (if not thrown) otherwise filtered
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -487,11 +487,8 @@ private void publishExceptionMetric(final Action action, final Throwable ex, fin
     private void
         publishExceptionCodeAndCountMetric(final Action action, final HandlerErrorCode handlerErrorCode, final boolean thrown) {
         if (this.metricsPublisherProxy != null) {
-            EnumSet.allOf(HandlerErrorCode.class).stream()
-                // publishing 0 value for all (if not thrown) otherwise filtered
-                .filter(errorCode -> errorCode.equals(handlerErrorCode) || !thrown)
-                .forEach(errorCode -> this.metricsPublisherProxy.publishExceptionByErrorCodeMetric(Instant.now(), action,
-                    errorCode, thrown));
+            EnumSet.allOf(HandlerErrorCode.class).forEach(errorCode -> this.metricsPublisherProxy
+                .publishExceptionByErrorCodeMetric(Instant.now(), action, errorCode, thrown && errorCode == handlerErrorCode));
```
wbingli, line 491: We should publish all those ErrorCode metrics in one PutMetric call, which can support multiple metrics in one request.. Otherwise it will consume large number of TPS, and may cause throttling.

PR #308, opened 2020-09-11
Alternate entry point without Lambda

ninavolkmuth:
*Description of changes:* This PR introduces another entry point to the plugin that completely removes the lambda dependency. This allows handler code to run on any platform and will not be limited to just AWS Lambda environments. 

To do this, a new parent wrapper class was introduced that holds all the core logic to handle CRUDL requests. This new parent class abstracts away the lambda logger into a generic log publisher so that both the lambda wrapper and the new executable wrapper could use their own loggers. Changes to the cloudwatch metrics and log helper classes were also made in order to use the generic log publisher. 

Diff for the wrapper: https://www.textcompare.org/index.html?id=5f68caea124d230017b1d14f

There is a new child class (ExecutableWrapper.java) that serves as the base for the entry point without the lambda dependency executable wrapper logs to both a file and the console. Both this new child class and the lambda wrapper class call the base wrapper's processRequest method after setting up the platform logger. 

The last changes are to the codegen. It adds logging configuration and will create a separate java class that serves as the alternate entry point. 

Tested these changes with a personal resource. 


python/rpdk/java/templates/generate/HandlerWrapper.java
```diff
@@ -108,8 +115,35 @@ public void testEntrypoint(
             response = ProgressEvent.defaultFailureHandler(e, HandlerErrorCode.InternalFailure);
         } finally {
             writeResponse(outputStream, response);
+            outputStream.close();
+        }
+    }
+    {% else %}
+    public static void main(String[] args) throws IOException {
+        if (args.length != 1){
```
jotompki, line 123: one arg is just the stringified handler request?

python/rpdk/java/templates/init/shared/log4j2.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jotompki, line 1: this can be in the data directory and just copied in. no need to render a template with no variables

python/rpdk/java/templates/generate/HandlerWrapper.java
```diff
@@ -108,8 +115,35 @@ public void testEntrypoint(
...
+            System.exit(1);
+        }
+        final String outputFile = (UUID.randomUUID().toString() + ".txt");
```
jotompki, line 126: do we need to keep track of this output? or are we just getting the output from system out?

python/rpdk/java/templates/generate/HandlerWrapper.java
```diff
@@ -108,8 +115,35 @@ public void testEntrypoint(
...
+        try(FileOutputStream output = new FileOutputStream(outputFile)){
+            try(InputStream input=IOUtils.toInputStream(args[0],"UTF-8")){
+                new ExecutableHandlerWrapper().handleRequest(input, output);
+                output.flush();
+            }
```
jotompki, line 131: is there a way to print out the output using the output stream instead of saving the file then reading it again?

jotompki (commented):
Didn't go into the Wrapper.java class yet just an initial go through - are there any large changes from the lambdawrapper class or is it just moving the contents into a new class for inheritance purposes?

python/rpdk/java/templates/generate/HandlerWrapper.java
```diff
@@ -79,6 +83,9 @@ private void initialiseHandlers() {
         return result;
     }
 
+    {% if wrapper_parent == "LambdaWrapper" -%}
```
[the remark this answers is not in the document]
ninavolkmuth: Yea that would be good. I know there was some confusion before about the fact that they behaved a little differently

python/rpdk/java/templates/generate/HandlerWrapper.java
ninavolkmuth: Yes

python/rpdk/java/templates/init/shared/log4j2.xml
ninavolkmuth: Good point, will do

python/rpdk/java/templates/generate/HandlerWrapper.java
ninavolkmuth: No, we are just getting the output from system out

python/rpdk/java/templates/generate/HandlerWrapper.java
ninavolkmuth: To get the results from a jar, you use either system out or can read from a file that the jar wrote to. I chose system out because that doesn't require the consumer to also read from a file (which also includes mounting a volume in an image so that they can access the file from the container as well). However, standard out also includes the logs so I need some way to differentiate the logs and the handler result. I can't just split it by newline and take the last element in the list because there may be new lines in the actual handler result. I ended up doing this way in order to be able to use a regex in order to grab the response in the cfn cli

ninavolkmuth:
> Didn't go into the Wrapper.java class yet just an initial go through - are there any large changes from the lambdawrapper class or is it just moving the contents into a new class for inheritance purposes?

Updated the description with a link to the diff

python/rpdk/java/templates/generate/HandlerWrapper.java
```diff
@@ -62,11 +67,10 @@ private void initialiseHandlers() {
 
     @Override
     public ProgressEvent<{{ pojo_name }}, CallbackContext> invokeHandler(
-                final AmazonWebServicesClientProxy proxy,
-                final ResourceHandlerRequest<{{ pojo_name }}> request,
-                final Action action,
-                final CallbackContext callbackContext) {
-
+            final AmazonWebServicesClientProxy proxy,
```
wbingli, line 70: is only format change here?

python/rpdk/java/codegen.py
```diff
@@ -304,6 +315,9 @@ def _init_settings(self, project):
         project.runtime = self.RUNTIME
         project.entrypoint = self.ENTRY_POINT.format(self.package_name)
         project.test_entrypoint = self.TEST_ENTRY_POINT.format(self.package_name)
+        project.executable_entrypoint = self.EXECUTABLE_ENTRY_POINT.format(
```
wbingli, line 318: We should also add the `executable_entrypoint` configuration in the `generate` command if it doesn't exist (but don't override it).

python/rpdk/java/templates/generate/HandlerWrapper.java
```diff
@@ -108,8 +115,35 @@ public void testEntrypoint(
...
+        readFileToSystemOut(outputFile);
+    }
+
+    private static void readFileToSystemOut(final String fileName) throws IOException {
+        //Create object of FileReader
+        final FileReader inputFile = new FileReader(fileName);
+        try(BufferedReader bufferReader = new BufferedReader(inputFile)) {
+            String line;
+            while ((line = bufferReader.readLine()) != null)   {
+                System.out.println(line);
```
wbingli, line 142: What if the handler using the stdout to print something in the middle? Will it impact the result of handler execution?

If so, you mostly need to put the resource with [some boundary lines](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html) like?
```
==Cloudformation resource result, boundary=boundary{UUID}
....
--boundary{UUID}-- 

```

src/main/java/software/amazon/cloudformation/Wrapper.java
```diff
@@ -0,0 +1,567 @@
...
+import software.amazon.cloudformation.proxy.ProgressEvent;
+import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
+import software.amazon.cloudformation.proxy.WaitStrategy;
+import software.amazon.cloudformation.resource.ResourceTypeSchema;
+import software.amazon.cloudformation.resource.SchemaValidator;
+import software.amazon.cloudformation.resource.Serializer;
+import software.amazon.cloudformation.resource.Validator;
+import software.amazon.cloudformation.resource.exceptions.ValidationException;
+
+public abstract class Wrapper<ResourceT, CallbackT> {
```
wbingli, line 72: Could we name it as `AbstractWrapper` to avoid misuse it.

src/main/java/software/amazon/cloudformation/Wrapper.java
```diff
@@ -0,0 +1,567 @@
...
+    /*
+     * This .ctor provided for testing
+     */
+    public Wrapper(final CredentialsProvider providerCredentialsProvider,
+                   final LogPublisher platformEventsLogger,
+                   final CloudWatchLogPublisher providerEventsLogger,
+                   final MetricsPublisher providerMetricsPublisher,
+                   final SchemaValidator validator,
+                   final Serializer serializer,
+                   final SdkHttpClient httpClient) {
```
wbingli, line 117: Why we need to add this `SdkHttpClient` as one of Wrapper constructor?   We are using a shared the HTTP client for all client to share connection pool etc.  It should be just to use a default static HTTP client.

src/main/java/software/amazon/cloudformation/Wrapper.java
```diff
@@ -0,0 +1,567 @@
...
+            if (request != null && request.getRequestData() != null && MUTATING_ACTIONS.contains(request.getAction())) {
+                handlerResponse.setResourceModel(request.getRequestData().getResourceProperties());
+            }
+            if (request != null) {
+                publishExceptionMetric(request.getAction(), e, HandlerErrorCode.InternalFailure);
+            }
+
+        } finally {
+            if (handlerResponse != null) {
+                publishExceptionCodeAndCountMetric(request == null ? null : request.getAction(), handlerResponse.getErrorCode(),
```
wbingli, line 226: There is more new changes on the LabmdaWrapper. Make sure merge the latest change after the refactoring, otherwise it will cause regression.   A recent PR will have new changes on this metric publish.

wbingli (commented):
Generally looks good.   

Other question is about backward compatibility on the PyPi and java maven version bump. A common scenario is:
 - What if customer upgrade the `cloudformation-cli` PyPi only? Will it break existing build?
 - What if customer upgrade the `cloudformation-cli-java-plugin` PyPI only without upgrading maven version? will it break the build?
 - What if customer only upgrade `aws-cloudformation-rpdk-java-plugin` maven plugin? will it break anything?  They may use `<version>[2.0.0, 3.0.0)</version>` in the maven for always pull the latest version? Will it  need a major version bump?

python/rpdk/java/templates/generate/HandlerWrapper.java
ninavolkmuth: The reason why I write to a file and then read it is specifically so that the response will be easy to regex and there will not be any logs in between the start/end response tags. The last line will look like this:

```
StartResponse-{RESPONSE_HERE}-EndResponse
```

I don't think boundary lines with a UUID will benefit this further

src/main/java/software/amazon/cloudformation/Wrapper.java
ninavolkmuth: That is a much better name. Will use that instead

src/main/java/software/amazon/cloudformation/Wrapper.java
ninavolkmuth: I did not touch this code, I just copied it from the existing lambdaWrapper. From what I could tell, this was done only so that we could mock the httpClient during testing easily

src/main/java/software/amazon/cloudformation/Wrapper.java
ninavolkmuth: Since I'm refactoring, any changes to the LambdaWrapper will cause merge conflicts so I will do this at the end and make sure the merge is also reviewed. I just don't want to keep fixing the merge conflicts while also making code changes so it's not as confusing looking at the diffs

python/rpdk/java/templates/generate/HandlerWrapper.java
ninavolkmuth: Yes. Just changed it to match the indentation level of other methods

python/rpdk/java/codegen.py
```diff
@@ -345,6 +359,19 @@ def generate(self, project):
             package_name=self.package_name,
             operations=project.schema.get("handlers", {}).keys(),
             pojo_name="ResourceModel",
+            wrapper_parent="LambdaWrapper",
+        )
+        project.overwrite(path, contents)
+
+        # write generated handler integration with ExecutableWrapper
+        path = src / "ExecutableHandlerWrapper.java"
```
wbingli, line 367: Generate this new class will cause build break if customer doesn't update the java plugin version.  And it happens quite common as update java plugin needs to update the pom.xml. 

One way we use during 2.0 upgrade is to check the java plugin version, and give an error if doesn't match.  I think you may can leverage the `_get_java_plugin_dependency_version` to check only generate this class if version is 2.1.0 or above. Assuming we will bump the java plugin version to 2.1.0.

python/rpdk/java/templates/generate/HandlerWrapper.java
wbingli: Got it. But do you think the boundary logic should be put here? The wrapper is the same logic to write the response to an output stream.

like 
```
System.out.println("__CFN_RESOURCE_START_RESPONSE__");
readFileToSystemOut(fileName);
System.out.println("__CFN_RESOURCE_END_RESPONSE__");
```

ninavolkmuth: Yea that makes sense to move it to the handlerwrapper class

python/rpdk/java/codegen.py
ninavolkmuth: Good call out. Current version is 2.0.2 so I just said anything greater than 2.0.3 so we can bump to that or 2.1

python/rpdk/java/codegen.py
```diff
@@ -403,6 +422,39 @@ def generate(self, project):
...
         LOG.debug("Generate complete")
 
+    def _write_executable_wrapper_class(self, src, project):
+        try:
+            java_plugin_dependency_version = self._get_java_plugin_dependency_version(
+                project
+            )
+            if (
+                java_plugin_dependency_version
+                >= MINIMUM_JAVA_DEPENDENCY_VERSION_EXECUTABLE_HANDLER_WRAPPER
```
wbingli, line 432: nice

python/rpdk/java/data/log4j2.xml
```diff
@@ -0,0 +1,17 @@
...
+    <Appenders>
+        <Console name="Console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+        </Console>
+        <File name="APPLICATION" fileName="log/application.log">
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+        </File>
+    </Appenders>
+    <Loggers>
+        <Root level="TRACE">
```
wbingli, line 12: nit:  `TRACE` level seems too fine grain. may be DEBUG level.

wbingli (approved):

jotompki (approved):

PR #310, opened 2020-09-15
added invalid request on invalid payload deserialization

ammokhov:
*Issue #, if available:*

*Description of changes:*

Currently, when we hit `MismatchedInputException` on payload deserialization we get `InternalFailure`, which is not surfacing real issue to the customer; Alternatively we could put model validation upfront but then we will have strict validation on stringification; this is a good alternative as customers can see that the input is actually invalid

### DIFF
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -210,7 +210,7 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
                 this.validator.validateObject(rawModelObject, resourceSchemaJSONObject);
 
                 handlerResponse = ProgressEvent.defaultFailureHandler(
-                    new CfnInvalidRequestException("Model validation failed caused by invalid input provided", e),
+                    new CfnInvalidRequestException("Resource properties validation failed with invalid configuration", e),
                     HandlerErrorCode.InvalidRequest);
             }
 
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -25,6 +25,8 @@ import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Arrays;
@@ -180,6 +182,15 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
         }
     }
 
+    private static String convertStackTraceToString(Throwable throwable) {
+        try (StringWriter sw = new StringWriter(); PrintWriter pw = new PrintWriter(sw)) {
+            throwable.printStackTrace(pw);
+            return sw.toString();
+        } catch (IOException ioe) {
+            throw new IllegalStateException(ioe);
+        }
+    }
+
     @Override
     public void handleRequest(final InputStream inputStream, final OutputStream outputStream, final Context context)
         throws IOException,
@@ -200,12 +211,16 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
             // deserialize incoming payload to modelled request
             try {
                 request = this.serializer.deserialize(input, typeReference);
+
+                handlerResponse = processInvocation(rawInput, request, context);
             } catch (MismatchedInputException e) {
                 JSONObject resourceSchemaJSONObject = provideResourceSchemaJSONObject();
                 JSONObject rawModelObject = rawInput.getJSONObject("requestData").getJSONObject("resourceProperties");
                 this.validator.validateObject(rawModelObject, resourceSchemaJSONObject);
+                handlerResponse = ProgressEvent.defaultFailureHandler(new TerminalException("#: Invalid input provided", e),
+                    HandlerErrorCode.InvalidRequest);
             }
-            handlerResponse = processInvocation(rawInput, request, context);
+
         } catch (final ValidationException e) {
             String message;
             String fullExceptionMessage = ValidationException.buildFullExceptionMessage(e);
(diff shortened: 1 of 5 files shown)


src/test/java/software/amazon/cloudformation/LambdaWrapperTest.java
```diff
@@ -332,7 +362,7 @@ public void invokeHandler_CompleteSynchronously_returnsSuccess(final String requ
         }
     }
 
-    @Test
+    // @Test
```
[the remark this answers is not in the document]
ammokhov: oops

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -180,6 +182,15 @@ private void initialiseRuntime(final String resourceType,
         }
     }
 
+    private static String convertStackTraceToString(Throwable throwable) {
```
wbingli, line 185: Can we use `ExceptionUtils` from apache common lang?  it has methods to convert as string.

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -200,12 +211,16 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
             // deserialize incoming payload to modelled request
             try {
                 request = this.serializer.deserialize(input, typeReference);
+
+                handlerResponse = processInvocation(rawInput, request, context);
             } catch (MismatchedInputException e) {
                 JSONObject resourceSchemaJSONObject = provideResourceSchemaJSONObject();
                 JSONObject rawModelObject = rawInput.getJSONObject("requestData").getJSONObject("resourceProperties");
                 this.validator.validateObject(rawModelObject, resourceSchemaJSONObject);
+                handlerResponse = ProgressEvent.defaultFailureHandler(new TerminalException("#: Invalid input provided", e),
```
wbingli, line 220: Why  using "#: " at the beginning of the error message?   Also TerminalException seems not right, how about CfnInvalidRequestException?

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -221,7 +236,7 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
         } catch (final Throwable e) {
             // Exceptions are wrapped as a consistent error response to the caller (i.e;
             // CloudFormation)
-            e.printStackTrace(); // for root causing - logs to LambdaLogger by default
+            log(convertStackTraceToString(e)); // for root causing - logs to LambdaLogger by default
```
wbingli, line 239: Nice

src/test/java/software/amazon/cloudformation/data/create.request.with-extraneous-model-object.json
```diff
@@ -0,0 +1,43 @@
...
+  "bearerToken": "123456",
+  "region": "us-east-1",
+  "action": "CREATE",
+  "responseEndpoint": "https://cloudformation.us-west-2.amazonaws.com",
+  "resourceType": "AWS::Test::TestModel",
+  "resourceTypeVersion": "1.0",
+  "requestContext": {},
+  "requestData": {
+    "callerCredentials": {
+      "accessKeyId": "IASAYK835GAIFHAHEI23",
```
wbingli, line 12: can you update the keys to a simple fake value?

src/test/java/software/amazon/cloudformation/LambdaWrapperTest.java
```diff
@@ -190,6 +190,36 @@ public void invokeHandler_nullResponse_returnsFailure(final String requestDataPa
...
+                any(HandlerErrorCode.class));
+
+            // all metrics should be published, even for a single invocation
+            verify(providerMetricsPublisher).publishInvocationMetric(any(Instant.class), eq(Action.CREATE));
+
+            // verify initialiseRuntime was called and initialised dependencies
+            verifyInitialiseRuntime();
+
+            // verify output response
+            verifyHandlerResponse(out, ProgressEvent.<TestModel, TestContext>builder().errorCode(HandlerErrorCode.InvalidRequest)
```
wbingli, line 218: The InvalidRequest looks right. But how does it match with previous `TerminalException`?

src/test/java/software/amazon/cloudformation/data/create.request.with-extraneous-model-object.json
```diff
@@ -0,0 +1,43 @@
...
+      "accessKeyId": "HDI0745692Y45IUTYR78",
+      "secretAccessKey": "4976TUYVI234/5GW87ERYG823RF87GY9EIUH452I3",
+      "sessionToken": "842HYOFIQAEUDF78R8T7IU43HSADYGIFHBJSDHFA87SDF9PYvN1CEYASDUYFT5TQ97YASIHUDFAIUEYRISDKJHFAYSUDTFSDFADS"
+    },
+    "providerLogGroupName": "providerLoggingGroupName",
+    "logicalResourceId": "myBucket",
+    "resourceProperties": {
+      "property1": "abc",
+      "property2": 123,
+      "fieldCausesValidationError": {
```
wbingli, line 26: So the deserialization will fail if nested properties? But how the validation can successful? There is `additionalProperties: false` for the schema, so additional property should fail on validation.

src/test/java/software/amazon/cloudformation/LambdaWrapperTest.java
ammokhov: it matches bc terminal exception was used just to create throwable but handler error code is set to `InvalidRequest`:

```java
ProgressEvent.defaultFailureHandler(
    new TerminalException("Model validation failed caused by invalid input provided", e),
    HandlerErrorCode.InvalidRequest);
```

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -200,12 +201,19 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
...
+
+                handlerResponse = processInvocation(rawInput, request, context);
             } catch (MismatchedInputException e) {
                 JSONObject resourceSchemaJSONObject = provideResourceSchemaJSONObject();
                 JSONObject rawModelObject = rawInput.getJSONObject("requestData").getJSONObject("resourceProperties");
+
                 this.validator.validateObject(rawModelObject, resourceSchemaJSONObject);
+
+                handlerResponse = ProgressEvent.defaultFailureHandler(
+                    new CfnInvalidRequestException("Model validation failed caused by invalid input provided", e),
```
wbingli, line 213: This message will be customer visible.  Maybe better wording as "Resource properties validation failed with invalid configuration".

src/test/resources/software/amazon/cloudformation/wrapper-override.json
```diff
@@ -1,12 +1,33 @@
 {
     "typeName": "Test::Resource::Type",
     "description": "Description",
+    "definitions": {
+        "subProperty": {
+            "type": "object",
+            "properties": {
```
wbingli, line 7: All right, I think the whole problem is that the schema doesn't define `"additionalProperties": false` for the nest property. In such way, the deserialization to POJO fails due to this mismatch. However, validation pass as `"additionalProperties": false` is missing and validation will skip this additional properties. 

The resource should fix the schema in such cases. But of course, this fix is still valuable.

wbingli (approved):
LGTM.  Better to update the error message as it will be customer visible.

sungkkim (approved):

PR #311, opened 2020-09-17
metrics refactor to publish in bulk

ammokhov:
*Issue #, if available:*

*Description of changes:*

refactoring publishing method: now accepts `MetricDatum`; metrics by error code are put together in bulk to avoid throttling possibility;

### DIFF
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -241,10 +241,8 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
             // A response will be output on all paths, though CloudFormation will
             // not block on invoking the handlers, but rather listen for callbacks
 
-            if (handlerResponse != null) {
-                publishExceptionCodeAndCountMetrics(request == null ? null : request.getAction(), handlerResponse.getErrorCode());
-            }
             writeResponse(outputStream, handlerResponse);
+            publishExceptionCodeAndCountMetrics(request == null ? null : request.getAction(), handlerResponse.getErrorCode());
         }
     }
 
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -29,7 +29,6 @@ import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Date;
-import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


src/main/java/software/amazon/cloudformation/metrics/MetricsPublisherImpl.java
```diff
@@ -57,86 +57,83 @@ public void publishExceptionMetric(final Instant timestamp,
...
+                .unit(StandardUnit.COUNT).value(errorCode == handlerErrorCode ? 1.0 : 0.0)
+                .dimensions(Sets.newHashSet(
+                    Dimension.builder().name(Metric.DIMENSION_KEY_ACTION_TYPE).value(action == null ? "NO_ACTION" : action.name())
+                        .build(),
+                    Dimension.builder().name(Metric.DIMENSION_KEY_HANDLER_ERROR_CODE).value(errorCode.name()).build()))
+                .timestamp(timestamp).build()));
+
+        // By Count dimensions
+        bulkData.add(MetricDatum.builder().metricName(Metric.METRIC_NAME_HANDLER_EXCEPTION_BY_EXCEPTION_COUNT)
+            .unit(StandardUnit.COUNT).value(1.0).dimensions(Dimension.builder().name(Metric.DIMENSION_KEY_ACTION_TYPE)
```
wbingli, line 90: This should not be `1.0` all the time, otherwise, the metrics will be `1` for every data point. 

It should be `handlerErrorCode == null ? 0.0 : 1.0` , if there is error code, the it's exception.

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -243,8 +242,7 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
             // not block on invoking the handlers, but rather listen for callbacks
 
             if (handlerResponse != null) {
-                publishExceptionCodeAndCountMetric(request == null ? null : request.getAction(), handlerResponse.getErrorCode(),
-                    handlerResponse.getStatus() == OperationStatus.FAILED);
+                publishExceptionCodeAndCountMetrics(request == null ? null : request.getAction(), handlerResponse.getErrorCode());
```
wbingli, line 245: Is that possbile we could move the `publishExceptionCodeAndCountMetrics` after the `writeResponse`? So that the lambda function should finish and publish metrics won't introduce additional latency to the handler execution.

wbingli (changes requested):

src/main/java/software/amazon/cloudformation/metrics/MetricsPublisherImpl.java
ammokhov: good catch!

wbingli (commented):
Look good to me.  Is there any test we have verified that moving the metric publish after write the response works?

ammokhov:
> Look good to me. Is there any test we have verified that moving the metric publish after write the response works?

since write response modifies value by reference it does not really affect metrics publishing unless write response has an issues itself; all the unit tests are passing + tested manually with different valid/invalid templates and all look good

wbingli (approved):

jotompki (approved):

Issue #299, opened 2020-08-19
Provide `StackId` in the ResourceHandlerRequest


wbingli:
StackId is useful information for the resource to get context. Also some resources uses stackId to generate name with stack name.

Current workaround is [using system tags](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/blob/master/aws-logs-loggroup/src/main/java/software/amazon/logs/loggroup/CreateHandler.java#L69). However, this information should be directly provided as public method in `ResourceHandlerRequest` to the handler.

ammokhov:
tracking #239 in here

PR #314, opened 2020-09-25
Provide stackId in ResourceHandlerRequest

omkhegde:
Issue #, if available: #299 

*Description of changes:*

### DIFF
--- a/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/LambdaWrapper.java
@@ -272,6 +272,7 @@ public abstract class LambdaWrapper<ResourceT, CallbackT> implements RequestStre
 
         if (resourceHandlerRequest != null) {
             resourceHandlerRequest.setPreviousResourceTags(getPreviousResourceTags(request));
+            resourceHandlerRequest.setStackId(getStackId(request));
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -570,6 +571,15 @@ protected void scrubFiles() {
         return previousResourceTags;
     }
 
+    @VisibleForTesting
+    protected String getStackId(final HandlerRequest<ResourceT, CallbackT> request) {
+        if (request != null) {
+            return request.getStackId();
+        }
+
+        return "";
```
ammokhov, line 580: null maybe?

ammokhov (changes requested):

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
ammokhov: should we add this to all plugins @wbingli ?

ammokhov (approved):

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
wbingli: Yes, we should

wbingli (approved):

PR #316, opened 2020-10-01
adding snapshot support

ammokhov:
*Issue #, if available:*

*Description of changes:*

adding support for snapshot metadata

### DIFF
--- a/python/rpdk/java/templates/generate/HandlerWrapper.java
+++ b/python/rpdk/java/templates/generate/HandlerWrapper.java
@@ -138,6 +138,7 @@ public final class HandlerWrapper extends LambdaWrapper<{{ pojo_name }}, Callbac
             .logicalResourceIdentifier(request.getRequestData().getLogicalResourceId())
             .nextToken(request.getNextToken())
             .region(request.getRegion())
+            .snapshotRequested(request.getSnapshotRequested())
             .awsPartition(PartitionMetadata.of(Region.of(request.getRegion())).id())
             .build();
     }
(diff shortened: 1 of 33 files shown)


aygold92 (approved):
is there a way to test this end to end before we submit it?

python/rpdk/java/templates/generate/HandlerWrapper.java
```diff
@@ -138,6 +138,7 @@ public JSONObject provideResourceSchemaJSONObject() {
             .logicalResourceIdentifier(request.getRequestData().getLogicalResourceId())
             .nextToken(request.getNextToken())
             .region(request.getRegion())
+            .snapshotRequested(request.getSnapshotRequested())
```
wbingli, line 141: We should not add change in the generated code here. This will cause build break if only upgrade the PyPi java plugin.

The better place like this https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/LambdaWrapper.java#L275

We won't even need tool release.

wbingli (changes requested):

src/main/java/software/amazon/cloudformation/proxy/HandlerRequest.java
```diff
@@ -34,5 +34,6 @@
     private RequestData<ResourceT> requestData;
```
jotompki, line 34: nit: should this be in request data instead? seems relatedto the actual request and not cloudformation metadata related info

aygold92: Good point, but TBH I don't really understand the distinction here anyways.  Like, why are providerCredentials, providerLogGroupName, and logicalResourceId part of requestData while callbackContext, accountId, bearerToken, and action are not?

wbingli (approved):

Issue #313, opened 2020-09-23
Provide resource name generation following current cloudformation naming pattern


wbingli:
Once with https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/299 fixed,  a new resource name generation to follow current cloud-formation name pattern,  `<StackName>-<ResourceName>-<RandomString>`

Customers depends on this name convention for access control etc.

PR #317, opened 2020-10-07
added stack-id in resource name generation

ammokhov:
*Issue #, if available:* #313 

*Description of changes:*

Added method to generate resource name with the following convention: `<StackName>-<ResourceName>-<Hash>`

StackName could be retrieved from `StackId` or from system tags -> this overload supports both - if `StackId` does not match arn pattern then method assumes it received clean stack name so it will use it directly;

**desired resource name**: `my-stack-name-rdsdbclusterparametergroup-19ok0k9pebcx6`
**new stack name**: `my-stack-name-rdsdbclusterparametergroup-dlo2atz4kirm`

### DIFF
--- a/src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
+++ b/src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
@@ -15,10 +15,8 @@
 package software.amazon.cloudformation.resource;
 
 import com.google.common.base.Splitter;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import java.util.Arrays;
-import java.util.List;
 import java.util.Random;
 import java.util.regex.Pattern;
 import org.apache.commons.lang3.RandomStringUtils;
@@ -82,16 +80,15 @@ public class IdentifierUtils {
                                                     final String clientRequestToken,
                                                     final int maxLength) {
 
-        String stackName = stackId;
-
-        if (isStackPortionValid(stackId)) {
-            List<String> stackParts = ImmutableList.copyOf(STACKID_SPLITTER.split(stackId));
-            stackName = stackParts.get(1);
-        }
-
         if (maxLength < MIN_PHYSICAL_RESOURCE_ID_LENGTH) {
             throw new IllegalArgumentException("Cannot generate resource IDs shorter than " + MIN_PHYSICAL_RESOURCE_ID_LENGTH
                 + " characters.");
+        }
+
+        String stackName = stackId;
+
+        if (isStackArn(stackId)) {
+            stackName = STACKID_SPLITTER.splitToList(stackId).get(1);
         }
 
         // some services don't allow leading dashes. Since stack name is first, clean
@@ -125,7 +122,7 @@ public class IdentifierUtils {
         return IdentifierUtils.generateResourceIdentifier(prefix.toString(), clientRequestToken, maxLength);
     }
 
-    private static boolean isStackPortionValid(String stackIdPortion) {
+    private static boolean isStackArn(String stackIdPortion) {
         return STACK_PATTERN.matcher(stackIdPortion).matches() && Iterables.size(STACKID_SPLITTER.split(stackIdPortion)) == 3;
     }
 
--- a/src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
+++ b/src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
@@ -65,7 +65,7 @@ public class IdentifierUtils {
         generateResourceIdentifier(final String logicalResourceId, final String clientRequestToken, final int maxLength) {
         int maxLogicalIdLength = maxLength - (GUID_LENGTH + 1);
 
-        int endIndex = logicalResourceId.length() > maxLogicalIdLength ? maxLogicalIdLength : logicalResourceId.length();
+        int endIndex = Math.min(logicalResourceId.length(), maxLogicalIdLength);
 
         StringBuilder sb = new StringBuilder();
         if (endIndex > 0) {
@@ -106,7 +106,7 @@ public class IdentifierUtils {
         // for each dash separator (if needed). the rest if the characters
         // will get allocated evenly between the stack and resource names
 
-        final int freeCharacters = maxLength - 13 - (separate ? 2 : 0);
+        final int freeCharacters = maxLength - 13 - (separate ? 1 : 0);
         final int[] requestedLengths = new int[2];
 
         requestedLengths[0] = cleanStackName.length();
--- a/src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
+++ b/src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
@@ -28,7 +28,8 @@ public class IdentifierUtils {
     private static final int MIN_PHYSICAL_RESOURCE_ID_LENGTH = 15;
     private static final int MIN_PREFERRED_LENGTH = 17;
     private static final Splitter STACKID_SPLITTER = Splitter.on('/');
-    private static final Pattern STACK_PATTERN = Pattern.compile("^[a-z0-9-:]*stack/[-a-z0-9A-Z/]*");
+    private static final Pattern STACK_ARN_PATTERN = Pattern.compile("^[a-z0-9-:]*stack/[-a-z0-9A-Z/]*");
+    private static final Pattern STACK_NAME_PATTERN = Pattern.compile("^[-a-z0-9A-Z]*");
 
     private IdentifierUtils() {
     }
@@ -91,6 +92,10 @@ public class IdentifierUtils {
             stackName = STACKID_SPLITTER.splitToList(stackId).get(1);
         }
 
+        if (!isValidStackName(stackName)) {
+            throw new IllegalArgumentException(String.format("%s is not a valid Stack name", stackName));
+        }
+
         // some services don't allow leading dashes. Since stack name is first, clean
         // off any + no consecutive dashes
 
@@ -122,8 +127,12 @@ public class IdentifierUtils {
         return IdentifierUtils.generateResourceIdentifier(prefix.toString(), clientRequestToken, maxLength);
     }
 
-    private static boolean isStackArn(String stackIdPortion) {
-        return STACK_PATTERN.matcher(stackIdPortion).matches() && Iterables.size(STACKID_SPLITTER.split(stackIdPortion)) == 3;
+    private static boolean isStackArn(String stackId) {
+        return STACK_ARN_PATTERN.matcher(stackId).matches() && Iterables.size(STACKID_SPLITTER.split(stackId)) == 3;
+    }
+
+    private static boolean isValidStackName(String stackName) {
+        return STACK_NAME_PATTERN.matcher(stackName).matches();
     }
 
     private static int[] fairSplit(final int cap, final int[] buckets) {
--- a/src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
+++ b/src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
@@ -14,13 +14,23 @@
 */
 package software.amazon.cloudformation.resource;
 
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
```diff
@@ -66,4 +76,85 @@ public static String generateResourceIdentifier(final String logicalResourceId,
...
+                                                    final int maxLength) {
+
+        String stackName = stackId;
+
+        if (isStackPortionValid(stackId)) {
+            List<String> stackParts = ImmutableList.copyOf(STACKID_SPLITTER.split(stackId));
+            stackName = stackParts.get(1);
+        }
+
+        if (maxLength < MIN_PHYSICAL_RESOURCE_ID_LENGTH) {
```
wbingli, line 92: Is this current rule?

src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
```diff
@@ -66,4 +76,85 @@ public static String generateResourceIdentifier(final String logicalResourceId,
...
     }
+
+    public static String generateResourceIdentifier(final String stackId,
+                                                    final String logicalResourceId,
+                                                    final String clientRequestToken,
+                                                    final int maxLength) {
+
+        String stackName = stackId;
+
+        if (isStackPortionValid(stackId)) {
```
wbingli, line 87: `isStackPortionValid` seems wried name. how about `isStackArn`?

src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
```diff
@@ -66,4 +76,85 @@ public static String generateResourceIdentifier(final String logicalResourceId,
...
+            List<String> stackParts = ImmutableList.copyOf(STACKID_SPLITTER.split(stackId));
```
wbingli, line 88: Why need to copy to a ImmutableList `stackParts`? can be simple as `STACKID_SPLITTER.split(stackId)[1]`?

src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
wbingli: And if it's this rule, we should move it at the very beginning fo the method.

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,48 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
         // string that is left to fix the max length.
         assertThat(result).startsWith("my-re-");
     }
+
+    @Test
+    public void generateResourceIdentifier_withStackInvalidStackId() {
```
wbingli, line 53: why it's a InvalidStackId? I think you mean stackName? As long as you support it, it should not be invalid.

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,48 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
...
+
+        // to ensure randomness in the identity, the result will always be a random
+        // string PREFIXED by the size of
+        // string that is left to fix the max length.
+        assertThat(result).startsWith("my-m-");
+    }
+
+    @Test
+    public void generateResourceIdentifier_withStackInvalidStackIdPartitionMismatch() {
+        String result = IdentifierUtils.generateResourceIdentifier("stack/my-stack-name", "my-resource", "123456", 50);
```
wbingli, line 65: I don't think we should support it. CloudFormation stack name never have `/`.  If customer want to use their own prefix, they should not use this method. 

let's make it as fail to throw error if the stackId like this.

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,48 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
...
+
+    @Test
+    public void generateResourceIdentifier_withStackInvalidStackIdLessThanPreferredLen() {
+        String result = IdentifierUtils.generateResourceIdentifier("stack/my-stack-name", "my-resource", "123456", 16);
+        assertThat(result.length()).isLessThanOrEqualTo(16);
+
+        // to ensure randomness in the identity, the result will always be a random
+        // string PREFIXED by the size of
+        // string that is left to fix the max length.
+        assertThat(result).startsWith("stm-");
```
wbingli, line 82: I don't understand. where is the `stm` comes from?

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,48 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
...
+    @Test
+    public void generateResourceIdentifier_withStackValidStackId() {
+        try {
+            IdentifierUtils.generateResourceIdentifier(
+                "arn:aws:cloudformation:us-east-1:123456789012:stack/my-stack-name/084c0bd1-082b-11eb-afdc-0a2fadfa68a5",
+                "my-resource", "123456", 14);
+        } catch (IllegalArgumentException e) {
+            assertThat(e.getMessage()).isEqualTo("Cannot generate resource IDs shorter than 15 characters.");
+        }
+    }
```
wbingli, line 94: Could we add more test case for using stackId? 

Cases like :
- Both size can fit into max length
 - Very long stack name, but short logic resource id
 - Very long logic resource name, but short stack name
 - Both stack and resource name are very long

wbingli (changes requested):

src/main/java/software/amazon/cloudformation/resource/IdentifierUtils.java
ammokhov: yep

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
ammokhov: this is when u dont have enough length to fith both groups evenly and have a separator "-"

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
ammokhov: but this will never happen I guess, if stack arn is submitted then pattern will match otherwise linter will reject malformed stack names

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,84 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
...
+    public void generateResourceIdentifier_withStackNameBothFitMaxLen() {
+        String result = IdentifierUtils.generateResourceIdentifier(
+            "arn:aws:cloudformation:us-east-1:123456789012:stack/my-stack-name/084c0bd1-082b-11eb-afdc-0a2fadfa68a5",
+            "my-resource", "123456", 255);
+        assertThat(result.length()).isLessThanOrEqualTo(44);
+        assertThat(result).isEqualTo("my-stack-name-my-resource-hDoP0dahAFjd");
+    }
+
+    @Test
+    public void generateResourceIdentifier_withLongStackNameAndShotLogicalId() {
```
wbingli, line 95: my point of testing a long stack name, is that we need to verify the stack name is trimmed due to that long stack name. But the logical name is saved.  So you can have a max identifier as 36 or something to verify what's the expected behavior.

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,84 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
...
+            "my-very-very-very-very-very-very-long-custom-logical-id", "123456", 255);
+        assertThat(result.length()).isLessThanOrEqualTo(72);
+        assertThat(result).isEqualTo("abc-my-very-very-very-very-very-very-long-custom-logical-id-hDoP0dahAFjd");
+    }
+
+    @Test
+    public void generateResourceIdentifier_withLongStackNameAndLongLogicalId() {
+        String result = IdentifierUtils.generateResourceIdentifier(
+            "arn:aws:cloudformation:us-east-1:123456789012:stack/my-very-very-very-very-very-very-long-custom-stack-name/084c0bd1-082b-11eb-afdc-0a2fadfa68a5",
+            "my-very-very-very-very-very-very-long-custom-logical-id", "123456", 255);
```
wbingli, line 115: same, let's use a small max ID length, such as 36 to verify both stack name and logic ID are trimmed properly.

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,93 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
...
+            "my-very-very-very-very-very-very-long-custom-logical-id", "123456", 36);
+        assertThat(result.length()).isLessThanOrEqualTo(35);
+        assertThat(result).isEqualTo("abc-my-very-very-very--hDoP0dahAFjd");
+    }
+
+    @Test
+    public void generateResourceIdentifier_withLongStackNameAndLongLogicalId() {
+        String result = IdentifierUtils.generateResourceIdentifier(
+            "arn:aws:cloudformation:us-east-1:123456789012:stack/my-very-very-very-very-very-very-long-custom-stack-name/084c0bd1-082b-11eb-afdc-0a2fadfa68a5",
+            "my-very-very-very-very-very-very-long-custom-logical-id", "123456", 255);
```
wbingli, line 115: Can we update this test case as well?

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,93 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
...
+        assertThat(result.length()).isLessThanOrEqualTo(44);
+        assertThat(result).isEqualTo("my-stack-name-my-resource-hDoP0dahAFjd");
+    }
+
+    @Test
+    public void generateResourceIdentifier_withLongStackNameAndShotLogicalId() {
+        String result = IdentifierUtils.generateResourceIdentifier(
+            "arn:aws:cloudformation:us-east-1:123456789012:stack/my-very-very-very-very-very-very-long-custom-stack-name/084c0bd1-082b-11eb-afdc-0a2fadfa68a5",
+            "abc", "123456", 36);
+        assertThat(result.length()).isLessThanOrEqualTo(35);
```
wbingli, line 99: Why it is 35? should it be 36? if the max length is 36?

src/test/java/software/amazon/cloudformation/resource/IdentifierUtilsTest.java
```diff
@@ -48,4 +48,93 @@ public void generateResourceIdentifier_withShortLength_prefixTruncated() {
...
+        assertThat(result).startsWith("mym-");
+    }
+
+    @Test
+    public void generateResourceIdentifier_withStackNameBothFitMaxLen() {
+        String result = IdentifierUtils.generateResourceIdentifier(
+            "arn:aws:cloudformation:us-east-1:123456789012:stack/my-stack-name/084c0bd1-082b-11eb-afdc-0a2fadfa68a5",
+            "my-resource", "123456", 255);
+        assertThat(result.length()).isLessThanOrEqualTo(44);
+        assertThat(result).isEqualTo("my-stack-name-my-resource-hDoP0dahAFjd");
```
wbingli, line 91: I want to double check the random string generated, is it has both lower case and upper case? or just one of it?  It would be great to follow extract the same name pattern.

wbingli (approved):

sungkkim (approved):

Issue #319, opened 2020-10-22
Log response payload for better troubleshooting


wbingli:
Currently there is no logging for the response payload. It makes the support and troubleshooting extremely hard. 

We should log the response payload for better troubleshooting, following should do:
 - Log full payload (https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/953665c7e844615447e16840e8e93338068c1080/src/main/java/software/amazon/cloudformation/AbstractWrapper.java#L385) if status is not IN_PROGRESS. That means the response has been sanitized. No sensitive in the payload.
 - Only log response status if the status is IN_PROGRESS.
 - Include `stackId` and resource logical ID as context of the log.

PR #320, opened 2020-10-22
Log response payload for better troubleshooting

sungkkim:
*Issue #, if available:*
[319](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/319)

*Description of changes:*

This change adds logging of stack id, logical resource id and response payload to aid in troubleshooting.


src/main/java/software/amazon/cloudformation/AbstractWrapper.java
```diff
@@ -383,6 +385,12 @@ protected void writeResponse(final OutputStream outputStream, final ProgressEven
         }
 
         String output = this.serializer.serialize(response);
+        if (response.getStatus() != OperationStatus.IN_PROGRESS) {
+            // if status is not IN_PROGRESS, it means the response has been sanitized
+            final String logicalResourceId = getLogicalResourceId(request);
+            final String stackId = getStackId(request);
+            this.log(String.format("StackId=%s LogicalResourceId=%s Response=%s", stackId, logicalResourceId, output));
```
ammokhov, line 392: should we log whole request?

wbingli: Request is not sanitized. We potentially can add the request as well after sanitized the request (e.g. sanitized the model and also other metadata). 

But I think we need to be a little bit more careful to add it step by step.  Besides, we should add the request log at the beginning of the handler. 

I would prefer to skip the log request for this PR and we can add it further.

src/main/java/software/amazon/cloudformation/AbstractWrapper.java
```diff
@@ -372,7 +372,9 @@ public void processRequest(final InputStream inputStream, final OutputStream out
 
     }
 
-    protected void writeResponse(final OutputStream outputStream, final ProgressEvent<ResourceT, CallbackT> response)
+    protected void writeResponse(final OutputStream outputStream,
```
wbingli, line 375: This seems a break change for the generated code.   I would suggest you do following to keep backward compatible.  And also update the code generate to use new method. In such way,  it won't break old code if still use old plugin, but it just won't have additional logging data. 

```
protected void writeResponse(final OutputStream outputStream,final ProgressEvent<ResourceT, CallbackT> response) {
 this(outputStream, response, null);
}

wbingli (approved):
Thanks

ammokhov (approved):

PR #327, opened 2020-11-06
[codegen] adding best practice to update tags

ammokhov:
*Issue #, if available:*

*Description of changes:*

adding best practice to the codegen on how to handle tag update:

* Stack level tags might be undesired tags, hence it's a best practice to catch permission issues.
* Resource level tags is just another resource property, so if user wants to attach resource tags then we should enforce correct permission set

### DIFF
--- a/python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
+++ b/python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
@@ -12,9 +12,11 @@ import software.amazon.cloudformation.proxy.Logger;
 import software.amazon.cloudformation.proxy.ProgressEvent;
 import software.amazon.cloudformation.proxy.ProxyClient;
 import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
+import software.amazon.cloudformation.proxy.HandlerErrorCode;
 
 public class UpdateHandler extends BaseHandlerStd {
     private Logger logger;
+    private static final String ACCESS_DENIED_EXCEPTION_MESSAGE = "not authorized";
 
     protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
         final AmazonWebServicesClientProxy proxy,
@@ -24,6 +26,8 @@ public class UpdateHandler extends BaseHandlerStd {
         final Logger logger) {
 
         this.logger = logger;
+
+        final ResourceModel previousModel = request.getPreviousResourceState();
 
         // TODO: Adjust Progress Chain according to your implementation
         // https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/proxy/CallChain.java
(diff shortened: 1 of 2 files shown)


sungkkim:
I can see the why we'd want to do this for backwards compatibility of existing resources, but do we really want to enforce this on newly introduced resources? Also, Create handler will have the same issue and if we go the same route, there could be issues with tag based access control. I'd like to hear some others' opinions as well.

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -134,7 +138,84 @@
...
+
+                // STEP 4.3 [TODO: check if event 1) failed 2) error code is access denied]
+                // if (event.isFailed() && ...) {
+                //     return progress;
+                // }
+                return event;
+            })
+
+            // If your resource supports tags, then the following pattern is required to handle resource level tags
+            // STEP 5 [update resource level tags progress chain]
```
wbingli, line 184: I'm confused. Is the Step 4 and 5  are the same?

ammokhov: step 4 has different translate request method - uses only stack tags:
```java
(model) -> Translator.translateToStackTagUpdateRequest(request.getPreviousResourceTags(), request.getDesiredResourceTags())
``` 

Step 5 has normal request translator - involves previous and current model
```java
.translateToServiceRequest((model) -> Translator.translateToResourceTagUpdateRequest(previousModel, model))
```

python/rpdk/java/templates/init/guided_aws/StubUpdateHandler.java
```diff
@@ -134,7 +138,84 @@
...
+                            // TODO: put your post update resource code here
+
+                        } catch (final AwsServiceException e) {
+                            /*
+                            * While the handler contract states that the handler must always return a progress event,
+                            * you may throw any instance of BaseHandlerException, as the wrapper map it to a progress event.
+                            * Each BaseHandlerException maps to a specific error code, and you should map service exceptions as closely as possible
+                            * to more specific error codes
+                            */
+                            throw new CfnGeneralServiceException(ResourceModel.TYPE_NAME, e);
```
wbingli, line 168: Then why this is a soft-failing?

ammokhov: stack level tags are not always intentional attachment by the customer so we dont really wanna enforce additional permissions when stack has not been changed on update

ninavolkmuth:
> I can see the why we'd want to do this for backwards compatibility of existing resources, but do we really want to enforce this on newly introduced resources? Also, Create handler will have the same issue and if we go the same route, there could be issues with tag based access control. I'd like to hear some others' opinions as well.

Yea, I don't think this will be solving the problem of existing resources adding tagging in a backwards incompatible manner. This code will have already been deleted for resources that are adding tagging later. We would also need to do it for create as well to cover replacement update operations

wbingli:
I agree this additional step is confusing to understand and I don't think it's a best practice for resource especially for any new resource and we don't want to encourage to ignore the stack level tag update.

Adding a new comments to mention to handler stack level tag properly would be better than this.  We should leverage other tools to prevent this happen instead of here.

PR #328, opened 2020-11-10
Long addition

jotompki:
*Issue #, if available:* #151, https://github.com/aws-cloudformation/cloudformation-cli/pull/621

*Description of changes:* Makes use of the additions in https://github.com/aws-cloudformation/cloudformation-cli/pull/621 to  generate "Long" primitive types when "int64" is specified as the integer format. Users can specify:

```
{
    "type": "integer",
    "format": "int64"
}
```
in their schema to generate a Long primitive in java. I also restructured the primitive types mapping to allow for additions of more formats in the future, if we want to support some natively supported jsonschema [string types](https://json-schema.org/understanding-json-schema/reference/string.html#built-in-formats) 

### DIFF
--- a/python/rpdk/java/resolver.py
+++ b/python/rpdk/java/resolver.py
@@ -1,4 +1,8 @@
-from rpdk.core.jsonutils.resolver import UNDEFINED, ContainerType
+import logging
+
+from rpdk.core.jsonutils.resolver import DEFAULT, UNDEFINED, ContainerType
+
+LOG = logging.getLogger(__name__)
 
 PRIMITIVE_TYPES = {
     "string": {"default": "String"},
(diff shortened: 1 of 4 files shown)


python/rpdk/java/resolver.py
```diff
@@ -1,27 +1,27 @@
...
+    UNDEFINED: {"default": "Object"},
 }
 
 
 def translate_type(resolved_type):
     if resolved_type.container == ContainerType.MODEL:
         return resolved_type.type
     if resolved_type.container == ContainerType.PRIMITIVE:
-        return PRIMITIVE_TYPES[resolved_type.type]
+        return PRIMITIVE_TYPES[resolved_type.type].get(resolved_type.type_format)
```
ammokhov, line 16: what if `type_format` is not in the `dict` then we get `None`

ammokhov (changes requested):

wbingli:
Is this PR still draft?

python/rpdk/java/resolver.py
jotompki: good call. will handle

jotompki:
> Is this PR still draft?

Since we are waiting on https://github.com/aws-cloudformation/cloudformation-cli/pull/621 to be merged and released for the functionality to actually work, I marked it as a draft until that work is done and that PR is merged and released to PyPI

jotompki:
Tested a newly registered resource with long types in a cloudformation stack. Values are passed to the handler as expected and users can !Ref a primary identifier that is a long, however attempting to perform !GetAtt on a readOnly Long property in the `Outputs` section results in a rollback. Confirmed this problem is present with integers as well, so we aren't introducing any errors by introducing the `Long` type

setup.py
```diff
@@ -36,7 +36,7 @@ def find_version(*file_paths):
     # package_data -> use MANIFEST.in instead
     include_package_data=True,
     zip_safe=True,
-    install_requires=["cloudformation-cli>=0.1.4,<0.2"],
+    install_requires=["cloudformation-cli>=0.1.14,<0.2"],
```
wbingli, line 39: Nice!

wbingli (approved):

ammokhov (approved):

jotompki:
Need to fix up the readme then I'll merge

PR #329, opened 2020-11-10
converter utility to cast multi-typed properties

ammokhov:
*Issue #, if available:*

*Description of changes:*

adding a utility class that helps to cast properties, which support multiple `type`/`$ref`, that were initially de serialized as `Object` types 

### DIFF
--- a/src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
+++ b/src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
@@ -23,18 +23,27 @@ public class TypeConverter {
         throw new UnsupportedOperationException();
     }
 
-    public static <T> Object covertProperty(final Object input, final TypeReference<?>... typeReferences) // preserves order
-        throws IOException {
-        final Serializer serializer = new Serializer();
-        final String stringified_input = serializer.serialize(input);
+    public static Object covertProperty(final Object input, final TypeReference<?>... typeReferences) // preserves order
+    {
+        try {
+            final Serializer serializer = new Serializer();
+            final String stringified_input = serializer.serialize(input);
 
-        for (TypeReference<?> typeReference : typeReferences) {
-            try {
-                return serializer.deserialize(stringified_input, typeReference);
-            } catch (JsonProcessingException ignored) {
-                // An empty catch block
+            int iter = 0;
+            while (true) {
+                TypeReference<?> typeReference = typeReferences[iter];
+                try {
+                    return serializer.deserialize(stringified_input, typeReference);
+                } catch (JsonProcessingException exception) {
+                    if (iter < typeReferences.length - 1) {
+                        iter++;
+                    } else {
+                        throw new RuntimeException("No Suitable Type Reference");
+                    }
+                }
             }
+        } catch (IOException exception) {
+            throw new RuntimeException("Invalid Object to Stringify");
         }
-        return input; // if no suitable type references were provided we dont wanna fail
     }
 }
--- a/src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
+++ b/src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
@@ -1,0 +1,40 @@
+/*
+* Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License").
+* You may not use this file except in compliance with the License.
+* A copy of the License is located at
+*
+*  http://aws.amazon.com/apache2.0
+*
+* or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.io.IOException;
+
+public class TypeConverter {
+    protected TypeConverter() {
+        throw new UnsupportedOperationException();
+    }
+
+    public static <T> Object covertProperty(final Object input, final TypeReference<?>... typeReferences) // preserves order
+        throws IOException {
```
wbingli, line 27: where is the IOException comes from?    I would suggest to catch it and throw with a RuntimeException.  The checked exception is not suitable here as it's not likely happen as all thing happen in memory.

src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+        final Serializer serializer = new Serializer();
+        final String stringified_input = serializer.serialize(input);
+
+        for (TypeReference<?> typeReference : typeReferences) {
+            try {
+                return serializer.deserialize(stringified_input, typeReference);
+            } catch (JsonProcessingException ignored) {
+                // An empty catch block
```
wbingli, line 35: Swallow exception is always could be a red flag and hard for further debug.

src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+            }
+        }
+        return input; // if no suitable type references were provided we dont wanna fail
```
wbingli, line 38: I would suggest don't return input if no matched. They should give all possible types, even give an Object as last option.  We should throw with exception if all given type doesn't match.

src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.io.IOException;
+
+public class TypeConverter {
+    protected TypeConverter() {
+        throw new UnsupportedOperationException();
+    }
+
+    public static <T> Object covertProperty(final Object input, final TypeReference<?>... typeReferences) // preserves order
```
wbingli, line 26: Why need a <T> here if the response is an Object type.  the `<T>` can be removed.

src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
ammokhov: IO comes from stringifying the object but yea, we can catch it

wbingli (approved):

src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
```diff
@@ -0,0 +1,49 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.io.IOException;
+
+public class TypeConverter {
+    protected TypeConverter() {
+        throw new UnsupportedOperationException();
+    }
+
+    public static Object covertProperty(final Object input, final TypeReference<?>... typeReferences) // preserves order
```
jotompki, line 26: spelling
```suggestion
    public static Object convertProperty(final Object input, final TypeReference<?>... typeReferences) // preserves order
```

src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
```diff
@@ -0,0 +1,49 @@
...
+    {
+        try {
+            final Serializer serializer = new Serializer();
+            final String stringified_input = serializer.serialize(input);
+
+            int iter = 0;
+            while (true) {
+                TypeReference<?> typeReference = typeReferences[iter];
+                try {
+                    return serializer.deserialize(stringified_input, typeReference);
```
jotompki, line 36: what happens if a multi-typed property has types string and int? would the ordering of how we specify the potential references matter? (not sure if this is the case) if a type correctly deserializes under both typeReferences passed into this function, why take the first over the second?

jotompki (changes requested):

src/main/java/software/amazon/cloudformation/resource/TypeConverter.java
ammokhov: yep the idea is to preserve the order

jotompki (approved):

PR #330, opened 2020-11-11
Skip validation on delete requests #318

jotompki:
*Issue #, if available:* #318 

*Description of changes:* Certain properties that are marked writeOnly and required will cause validation errors on delete, as the primaryIdentifier is only passed to the delete handlers. This PR removes the requirement to validate incoming payloads on deletion.


wbingli (commented):
Is this only for contract test?  I believe in Cloudformation a full resource model is passed for the delete handler.

jotompki:
> Is this only for contract test? I believe in Cloudformation a full resource model is passed for the delete handler.

Let me double check on this assumption and get back to you.

wbingli:
Even CFN pass the full model to the DELETE, it doesn't make sense to make validation for the resource model, especially for the writeOnly properties. It only makes sense to do a full model validation for CREATE/UPDATE.

wbingli (approved):

anshikg (approved):

ammokhov (approved):

ammokhov:
Did you run any manual tests? Did you make sure it does not cause any regression?

Issue #176, opened 2019-10-24
injectCredentialsAndInvoke doesn't support void returns.


rjlohan:
e.g; an API call to `setBucketNotificationConfiguration` for S3 returns void, hence the proxy call cannot support this API.

jotompki:
Should just need to add a void version of the method here: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java?

ammokhov:
so since aws sdk v2 has different model it always returns a response, so the change needs to be made in here - https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java#L439-L455

PR #338, opened 2020-12-08
Handle void response in injectCredentialsAndInvoke

omkhegde:
*Issue #, if available:*
#176 

*Description of changes:*

### DIFF
--- a/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
+++ b/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
@@ -27,6 +27,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
@@ -444,11 +445,24 @@ public class AmazonWebServicesClientProxy implements CallChain {
 
         try {
             ResultT response = requestFunction.apply(request);
-            if(response != null) {
-                logRequestMetadata(request, response);
-            }
-
+            logRequestMetadata(request, response);
             return response;
+        } catch (final Throwable e) {
+            loggerProxy.log(String.format("Failed to execute remote function: {%s}", e.getMessage()));
+            throw e;
+        } finally {
+            request.setRequestCredentialsProvider(null);
+        }
+    }
+
+    public <RequestT extends AmazonWebServiceRequest>
+    void
+    injectCredentialsAndInvoke(final RequestT request, final Consumer<RequestT> requestFunction) {
+
+        request.setRequestCredentialsProvider(v1CredentialsProvider);
+
+        try {
+            requestFunction.accept(request);
         } catch (final Throwable e) {
             loggerProxy.log(String.format("Failed to execute remote function: {%s}", e.getMessage()));
             throw e;
@@ -476,6 +490,8 @@ public class AmazonWebServicesClientProxy implements CallChain {
             throw e;
         }
     }
+
+
 
     public <RequestT extends AwsRequest, ResultT extends AwsResponse>
         CompletableFuture<ResultT>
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
```diff
@@ -443,9 +443,12 @@ public Logger getLogger() {
...
         try {
-            ResultT respose = requestFunction.apply(request);
-            logRequestMetadata(request, respose);
-            return respose;
+            ResultT response = requestFunction.apply(request);
+            if(response != null) {
+                logRequestMetadata(request, response);
+            }
+
+            return response;
```
ammokhov, line 451: will that work? dont you get an exception if you trying to assign to a value void return? this wont give you exception on the compilation time but might error out on runtime

src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
```diff
@@ -443,9 +444,24 @@ public Logger getLogger() {
...
+        } catch (final Throwable e) {
+            loggerProxy.log(String.format("Failed to execute remote function: {%s}", e.getMessage()));
+            throw e;
+        } finally {
+            request.setRequestCredentialsProvider(null);
+        }
+    }
+
+    public <RequestT extends AmazonWebServiceRequest> void injectCredentialsAndInvoke(final RequestT request,
+                                                                                      final Consumer<RequestT> requestFunction) {
```
ammokhov, line 459: so `Consumer<RequestT>` is an alternative for void functions?

src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
```diff
@@ -443,9 +444,24 @@ public Logger getLogger() {
...
         try {
-            ResultT respose = requestFunction.apply(request);
-            logRequestMetadata(request, respose);
-            return respose;
+            ResultT response = requestFunction.apply(request);
+            logRequestMetadata(request, response);
+            return response;
+        } catch (final Throwable e) {
+            loggerProxy.log(String.format("Failed to execute remote function: {%s}", e.getMessage()));
+            throw e;
```
ammokhov, line 452: do we want to wrap it in CFN exception? `CfnInvalidRequest` maybe - @wbingli

ammokhov (commented):
lgtm, just a couple of comments

src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
```diff
@@ -443,9 +444,24 @@ public Logger getLogger() {
         request.setRequestCredentialsProvider(v1CredentialsProvider);
 
         try {
-            ResultT respose = requestFunction.apply(request);
-            logRequestMetadata(request, respose);
-            return respose;
+            ResultT response = requestFunction.apply(request);
+            logRequestMetadata(request, response);
+            return response;
```
jotompki, line 449: nice typo fix

src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
jotompki: i think we just want to return whatever the sdk returns here so the provider knows what the exception of the sdk was - Cfn prefixed exceptions are for the providers to determine how they are applied. 

If you look down in the diff this is what we were doing previously

src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
jotompki, line 459: we should also just make a void function for the v2 proxy as well? if we are tackling the v1 here?

jotompki (approved):

src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
omkhegde: I think v2 functions are required to return a result and are never void. @ammokhov can you please confirm?

src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java
ammokhov: @jotompki agreed, the mapping will happen inside call chain

ammokhov (approved):

PR #339, opened 2020-12-17
GitHub Actions CI

jotompki:
*Issue #, if available:*

*Description of changes:* First two commits: Github actions CI for java. Should have parity with existing travis job.

Next commit: Adding a PyPI deployment actions. Any time a release w tag suffix "-plugin" is published, the CI job will publish said release to PyPI.

Final commit: Adds automation to release to our staging repository for releases w tag suffix "-lib". This should make it a lot easier to cut new releases, as all that is necessary is a version bump an a tag. Also adds a small configuration change to the pom.xml to support the automated deployment. 

For both of these, I added the credentials for uploading, keys, passphrases, etc. As GitHub secrets, which are encrypted and stored on an organization level. 


pom.xml
```diff
@@ -431,6 +431,12 @@
                                 <goals>
                                     <goal>sign</goal>
                                 </goals>
+                                <configuration>
```
jotompki, line 434: Had to add this to pick up the passphrase for the gpg key that is set in the setup java step

PatMyron:
Can we remove `.travis.yml` in these PRs as well? Easier to review the diff for parity if that's part of the PR

jotompki:
> Can we remove [`.travis.yml`](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/.travis.yml) in these PRs as well? Easier to review the diff for parity if that's part of the PR

Should be removed now.

ammokhov (approved):

.github/workflows/maven-release.yaml
```diff
@@ -0,0 +1,31 @@
...
+on:
+  release:
+    types: [ published ]
+
+jobs:
+  build:
+    if: endsWith(github.ref, "-lib")
+    env:
+      AWS_DEFAULT_REGION: us-east-1
+      AWS_REGION: us-east-1
```
PatMyron, line 13: Are these environment variables used for anything btw?
```suggestion
```

.github/workflows/pr-ci.yaml
```diff
@@ -0,0 +1,49 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    env:
+      AWS_DEFAULT_REGION: us-east-1
+      AWS_REGION: us-east-1
```
PatMyron, line 14: Are these environment variables used for anything btw?
```suggestion
```

.github/workflows/pr-ci.yaml
```diff
@@ -0,0 +1,49 @@
...
+    strategy:
+      matrix:
+        python: [3.6, 3.7, 3.8]
```
PatMyron, line 17: linking https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/issues/139 again
we should just sweep all these repos when we get Python 3.9 working:
```suggestion
        python: [3.6, 3.7, 3.8, 3.9]
```

.github/workflows/pr-ci.yaml
```diff
@@ -0,0 +1,49 @@
...
+        java: [8, 11]
+    runs-on: ubuntu-latest
```
PatMyron, line 19: might as well test on [`macos-latest` and `windows-latest`](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) here as well

.github/workflows/pypi-release.yaml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip wheel twine
```
PatMyron, line 20: Not sure if we need to `pip install pip` here:
```suggestion
        pip install --upgrade wheel twine
```

.github/workflows/maven-release.yaml
```diff
@@ -0,0 +1,31 @@
...
+        server-username: SONATYPE_USERNAME # env variable for username in deploy
+        server-password: SONATYPE_PASSWORD  # env variable for token in deploy
+        gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }} # Value of the GPG private key to import
+        gpg-passphrase: MAVEN_GPG_PASSPHRASE # env variable for GPG private key passphrase
+    - name: Deploy to sonatype staging repo
+      run: mvn deploy -Ppublishing
+      env:
+        SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+        SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+        MAVEN_GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
```
PatMyron, line 31: confused what's going on with [`secrets`](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets)/[`env variables`](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables) here

looks similar to [`actions/setup-java` README](https://github.com/actions/setup-java#publishing-using-apache-maven) but don't understand why `env` variables are set in the step after they're used 🤔

PatMyron (changes requested):

.github/workflows/maven-release.yaml
jotompki: yeah if you dont have these set the unit tests fail 🤷🏻 something with initializing aws clients in unit test

.github/workflows/maven-release.yaml
jotompki: i think it's to limit the scope of when these secrets are on the build machine. the  setup-java step simply creates a configuration file which points to environment variables then the deploy step sets these when the configuration file is actually used (during the `mvn deploy` command)

.github/workflows/pr-ci.yaml
jotompki: added. i think it might also be useful to test compatibility with published libraries and plugins i.e. since we are managing two different packages of software, we should test the library against the published  plugin and vice versa to ensure backwards compatibility. These things get missed since we test the built library in a given PR with the built plugin in the given PR

jotompki: seeing some macos failures trying to build the java library. going to leave out the multiple OS testing for now since this was just supposed to be a migration and create a followup so we can do this in a more targeted way where we aren't just running 18 jobs for one pr. ideally we'd only need to test compatibility of the plugin with the given OS

PatMyron (approved):

PR #349, opened 2021-01-21
Add plugin information in a json metadata file

omkhegde:
*Issue #, if available:*

*Description of changes:*
Adding plugin information in a metadata file named _cfn_metadata.json. This file will be included in the package artifact zip being uploaded. 

### DIFF
--- a/python/rpdk/java/codegen.py
+++ b/python/rpdk/java/codegen.py
@@ -49,8 +49,6 @@ DEFAULT_SETTINGS = {PROTOCOL_VERSION_SETTING: DEFAULT_PROTOCOL_VERSION}
 MINIMUM_JAVA_DEPENDENCY_VERSION = "2.0.0"
 MINIMUM_JAVA_DEPENDENCY_VERSION_EXECUTABLE_HANDLER_WRAPPER = "2.0.3"
 
-CFN_METADATA_FILE_NAME = "_cfn_metadata.json"
-
 
 class JavaArchiveNotFoundError(SysExitRecommendedError):
     pass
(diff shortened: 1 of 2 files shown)


python/rpdk/java/codegen.py
```diff
@@ -1,5 +1,6 @@
 # pylint: disable=useless-super-delegation,too-many-locals
 # pylint doesn't recognize abstract methods
+import json
```
omkhegde, line 3: No need to import json anymore

python/rpdk/java/codegen.py
```diff
@@ -571,3 +572,15 @@ def generate_image_build_config(self, project):
...
             "project_path": str(project_path),
         }
+
+    @staticmethod
+    def _get_plugin_information():
+        plugin_information = {}
+        plugin_information["plugin-version"] = __version__
+        plugin_information["plugin-name"] = "java"
+
+        return plugin_information
```
ammokhov, line 582: minor:
```python
return {
    "plugin-version": __version__,
    "plugin-name": "java",
}
```

ammokhov (approved):
Also make sure that cli or lib versions works with older versions, otherwise add install requirements

python/rpdk/java/codegen.py
```diff
@@ -571,3 +572,14 @@ def generate_image_build_config(self, project):
             "dockerfile_path": dockerfile_path,
             "project_path": str(project_path),
         }
+
+    @staticmethod
+    def _get_plugin_information():
+        return {
+            "plugin-version": __version__,
```
wbingli, line 579: This is collecting the plugin tool version.  We should also collect the plugin runtime version as well.  

We did similar plugin version check for 1.0 to 2.0 upgrade, there is utils method to retrieve the version. Although customer may not use pom.xml in the package, but it should be a best effort to collect the runtime version.

I think the metadata should like: 

```
{
"plugin-name": "java", 
"plugin-tool-version": "2.0.1",
"plugin-version": "2.0.3"

}
```

tests/test_codegen.py
```diff
@@ -344,3 +345,10 @@ def test_generate_executable_entrypoint_old_project_version(project):
     plugin._namespace_from_project(project)
 
     assert not hasattr(project, "executable_entrypoint")
+
+
+def test_get_plugin_information(project):
+    plugin_information = project._plugin.get_plugin_information(project)
+
+    assert plugin_information["plugin-tool-version"] == __version__
```
wbingli, line 353: any test to the `plugin-version` for the runtime?

wbingli (approved):

PR #364, opened 2021-04-06
add reason to CfnNotStabilizedException

cortez7:
The current stabilization error message "Resource of type '%s' with identifier '%s' did not stabilize." lacks a reason so users may not be able to determine the cause of the stabilization error.

### Changes
- add reason argument to CfnNotStabilizedException constructors with documentation
- update the corresponding NOT_STABILIZED exception message
- deprecate the old constructors

### Testing
- modified unit tests for new constructors

### DIFF
--- a/src/main/java/software/amazon/cloudformation/exceptions/CfnNotStabilizedException.java
+++ b/src/main/java/software/amazon/cloudformation/exceptions/CfnNotStabilizedException.java
@@ -25,14 +25,62 @@ public class CfnNotStabilizedException extends BaseHandlerException {
         super(cause, ERROR_CODE);
     }
 
+    /**
+     * @param resourceTypeName
+     * @param resourceIdentifier
+     * @param reason Reason why the resource did not stabilize. This should include the current and
+     *               desired state.
+     *
+     *               Example: "Exceeded retry limit for DescribeResourceStatus.
+     *               Current Value: IN_PROGRESS. Desired Value: COMPLETE."
+     */
     public CfnNotStabilizedException(final String resourceTypeName,
-                                     final String resourceIdentifier) {
-        this(resourceTypeName, resourceIdentifier, null);
+                                     final String resourceIdentifier,
+                                     final String reason) {
+        this(resourceTypeName, resourceIdentifier, reason, null);
     }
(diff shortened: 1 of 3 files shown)


src/main/java/software/amazon/cloudformation/exceptions/CfnNotStabilizedException.java
```diff
@@ -25,14 +25,62 @@ public CfnNotStabilizedException(final Throwable cause) {
...
     }
 
+    /**
+     * @param resourceTypeName
+     * @param resourceIdentifier
+     * @param reason Reason why the resource did not stabilize. This should include the current and
+     *               desired state.
+     *
+     *               Example: "Exceeded retry limit for DescribeResourceStatus.
+     *               Current Value: IN_PROGRESS. Desired Value: COMPLETE."
```
aygold92, line 35: nice, I like the docs and providing an example.  Couple comments:
* I don't think desired state is necessary
* For AWS resources, it should include the current state, Api name in iam format (like ec2:DescribeInstances) and the RequestId for that API (for non AWS resources it would be whatever the equivalent debugging info)

The example here is also a little bit confusing. Is this one example or two examples?  Here's one

```
* Instance Status: pending (Api: ec2:DescribeInstances, RequestId: abcd-efgh-ijkl)
* Current Status: modifying (Api: rds:DescribeDBClusters, RequestId: abcd-efgh-ijkl)
```

Another option here would be to take the apiName and requestid as parameters

aygold92 (changes requested):

src/main/java/software/amazon/cloudformation/exceptions/CfnNotStabilizedException.java
```diff
@@ -25,14 +25,64 @@ public CfnNotStabilizedException(final Throwable cause) {
         super(cause, ERROR_CODE);
     }
 
+    /**
+     * @param resourceTypeName
```
aygold92, line 29: nit: can remove this duplicate doc entry

aygold92 (approved):

sungkkim (approved):

PR #365, opened 2021-04-10
resolves #341

### LINKED ISSUE #341: [Proposal] Use unboxed boolean types in ResourceHandlerRequest
There are 2 recent changes that landed in `ResourceHandlerRequest`: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/commit/ebca1dc12e187b484ed599db65d726a6b23a14c0#diff-adfb207dac8dff5107b68b36280ccad4525f92176184ad2ffff9aa7e591b2e8a and https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/commit/e34f0472d0bc67953f41d4ceb666a175d0a7fb2d#diff-adfb207dac8dff5107b68b36280ccad4525f92176184ad2ffff9aa7e591b2e8a.

Both changes favor boxed Boolean over primitive types.

As a result, the usage of this class implies on importing BooleanUtils or a similar helper in order to avoid constructions like: `if (request.getSnapshotRequested() != null && request.getSnapshotRequested() == true) { ... }`.

And the code that is performing a "naive" boolean comparison like: `if (request.getSnapshotRequested()) { ... }` will fail with a null-pointer exception.

The same stands for the rollback flag.

I wonder if var type boxing was done for some specific reasons? If there are no strict reasons, I would vote for unboxing them to simplify condition checking.

pemmasanikrishna:
favor primitive types over boxed Boolean.

### DIFF
--- a/src/main/java/software/amazon/cloudformation/proxy/ResourceHandlerRequest.java
+++ b/src/main/java/software/amazon/cloudformation/proxy/ResourceHandlerRequest.java
@@ -43,9 +43,9 @@ public class ResourceHandlerRequest<T> {
     private String awsPartition;
     private String logicalResourceIdentifier;
     private String nextToken;
-    private Boolean snapshotRequested;
-    private Boolean rollback;
-    private Boolean driftable;
+    private boolean snapshotRequested;
+    private boolean rollback;
+    private boolean driftable;
     private String region;
     private String stackId;
 }
--- a/src/main/java/software/amazon/cloudformation/AbstractWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/AbstractWrapper.java
@@ -264,9 +264,9 @@ public abstract class AbstractWrapper<ResourceT, CallbackT> {
         if (resourceHandlerRequest != null) {
             resourceHandlerRequest.setPreviousResourceTags(getPreviousResourceTags(request));
             resourceHandlerRequest.setStackId(getStackId(request));
-            resourceHandlerRequest.setSnapshotRequested(request.getSnapshotRequested());
-            resourceHandlerRequest.setRollback(request.getRollback());
-            resourceHandlerRequest.setDriftable(request.getDriftable());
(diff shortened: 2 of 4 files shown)


ammokhov:
this seems to intentionally set `false`/`true` to the parameter that are initially set only in certain cases, otherwise `null`. @wbingli @PatMyron seeking ur opinions if we should generalize and always return boolean?

src/main/java/software/amazon/cloudformation/proxy/ResourceHandlerRequest.java
```diff
@@ -43,9 +43,9 @@
     private String awsPartition;
     private String logicalResourceIdentifier;
     private String nextToken;
-    private Boolean snapshotRequested;
-    private Boolean rollback;
-    private Boolean driftable;
+    private boolean snapshotRequested;
```
wbingli, line 46: I would suggest don't change the property type as it will cause method change and it will cause compile error.

As suggest in https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/341#issuecomment-818960517,   we can provide extra helper method, such as `isSnapshotRequested` to simplify it.

wbingli (changes requested):

src/main/java/software/amazon/cloudformation/proxy/ResourceHandlerRequest.java
pemmasanikrishna: @wbingli I have already taken care of that change and changed the references as well. And the pr got build is  successful. Do yoy want me to revert the change and provide helper method ?

wbingli: The point is that anyone who uses this class will be broken.  The java plugin can build successful doesn't mean it won't break others.  Therefore, I suggest to add helper method in the class instead of change the field type.

PR #401, opened 2022-07-22
CW client creds refresh

ammokhov:
*Issue #, if available:*

*Description of changes:*

### Change
this change refreshes provider creds when non-null during runtime initialization

### Testing

```
$ mvn verify --no-transfer-progress
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  49.317 s
[INFO] Finished at: 2022-07-22T11:22:20-07:00
[INFO] ------------------------------------------------------------------------
```

### DIFF
--- a/src/main/java/software/amazon/cloudformation/AbstractWrapper.java
+++ b/src/main/java/software/amazon/cloudformation/AbstractWrapper.java
@@ -88,8 +88,8 @@ public abstract class AbstractWrapper<ResourceT, CallbackT, ConfigurationT> {
     // provider... prefix indicates credential provided by resource owner
     protected final CredentialsProvider providerCredentialsProvider;
 
-    protected final CloudWatchProvider providerCloudWatchProvider;
-    protected final CloudWatchLogsProvider cloudWatchLogsProvider;
+    protected CloudWatchProvider providerCloudWatchProvider;
+    protected CloudWatchLogsProvider cloudWatchLogsProvider;
     protected final SchemaValidator validator;
     protected final TypeReference<HandlerRequest<ResourceT, CallbackT, ConfigurationT>> typeReference;
 
@@ -151,9 +151,13 @@ public abstract class AbstractWrapper<ResourceT, CallbackT, ConfigurationT> {
         // Both are required parameters when LoggingConfig (optional) is provided when
         // 'RegisterType'.
         if (providerCredentials != null) {
+
             if (this.providerCredentialsProvider != null) {
                 this.providerCredentialsProvider.setCredentials(providerCredentials);
             }
+
+            this.providerCloudWatchProvider = new CloudWatchProvider(this.providerCredentialsProvider, HTTP_CLIENT);
+            this.cloudWatchLogsProvider = new CloudWatchLogsProvider(this.providerCredentialsProvider, HTTP_CLIENT);
 
             if (this.providerMetricsPublisher == null) {
                 this.providerMetricsPublisher = new MetricsPublisherImpl(this.providerCloudWatchProvider, this.loggerProxy,


src/main/java/software/amazon/cloudformation/AbstractWrapper.java
```diff
@@ -151,10 +151,14 @@ public AbstractWrapper(final CredentialsProvider providerCredentialsProvider,
         // Both are required parameters when LoggingConfig (optional) is provided when
         // 'RegisterType'.
         if (providerCredentials != null) {
```
himanshs23, line 153: the only concern i think is, if we want to reset it to null,  we will not allow it right?

I can see three scenarios and curios on the third one
1. initially it was null, second request we got a non null provider credential we refresh it - it will work
2. initially it was non-null provider-credential, second request we got a new non-null provider credential - it will work for this case too
3. initially it was non-null provider-credenial, second request we got a null provider credential - We will still use the old one?

aygold92 (changes requested):
how was this tested?

aygold92 (approved):

src/main/java/software/amazon/cloudformation/LambdaWrapper.java
```diff
@@ -56,8 +56,8 @@ public void handleRequest(final InputStream inputStream, final OutputStream outp
         TerminalException {
         if (platformLogPublisher == null) {
             platformLogPublisher = new LambdaLogPublisher(context.getLogger());
+            this.platformLoggerProxy.addLogPublisher(platformLogPublisher);
         }
-        this.platformLoggerProxy.addLogPublisher(platformLogPublisher);
```
ammokhov, line 60: does it do much of difference? like we dont want to add publisher all the time?

src/main/java/software/amazon/cloudformation/AbstractWrapper.java
```diff
@@ -150,7 +150,15 @@ public AbstractWrapper(final CredentialsProvider providerCredentialsProvider,
...
-        if (providerCredentials != null) {
+        if (providerCredentials == null) {
+            // reset provider credentials back to null to avoid reusing stale credentials
+            if (this.providerCredentialsProvider != null) {
+                this.providerCredentialsProvider.resetCredentials();
+            }
+            this.providerMetricsPublisher = null;
+            this.providerEventsLogger = null;
+            this.cloudWatchLogHelper = null;
+        } else {
```
rajdivya, line 161: Why null check providerCredentials and providerCredentialsProvider before resetting the credentials? Cant we do it for every request? The next section will set it if credentials are provided anyway.

ammokhov: because it does get refreshed when creds are non-null

aygold92: Actually that null check on providerCredentialsProvider seemed unnecessary to me.  It's a final variable and instantiated in the constructor, but it was already there and I was just afraid to remove it and potentially cause an NPE for something I couldn't see.

rajdivya: Ok. What exactly are providerCredentials and providerCredentialsProvider? Do the credentials not need to be reset if both providerCredentials and providerCredentialsProvider are null?

aygold92: providerCredentials come directly from the request sent in by CloudFormation (or Cloud Control API).  So if the credentials passed by CFN are null, then we want to reset everything.  

ProviderCredentialsProvider is just a wrapper class which holds a reference to a credentials object.  Like I said I don't think it's possible to be null, but if it was then it wouldn't hold any credentials anyway, which is what we want in this case. And even then, we still reset the other related objects metricsPublisher, eventsLogger, and cloudWatchLogHelper

PR #419, opened 2023-04-26
Tagging Specific Error Code

ammokhov:
*Issue #, if available:*

*Description of changes:*

Adding a tagging specific error code. Decided to name it `UnauthorizedTaggingOperation` not to mix it with `AccessDenied`. Both are similar in nature but we need to be able to distinguish access denied caused by tagging operation;

### DIFF
--- a/src/main/java/software/amazon/cloudformation/exceptions/CfnUnauthorizedTaggingOperationException.java
+++ b/src/main/java/software/amazon/cloudformation/exceptions/CfnUnauthorizedTaggingOperationException.java
@@ -1,0 +1,31 @@
+/*
+* Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License").
+* You may not use this file except in compliance with the License.
+* A copy of the License is located at
+*
+*  http://aws.amazon.com/apache2.0
+*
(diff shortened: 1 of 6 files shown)


bcorman3 (approved):

src/main/java/software/amazon/cloudformation/exceptions/CfnUnauthorizedTaggingOperationException.java
shrsrivz, line 1: Can we follow the same convention to have overloaded constructors. For ex- CfnAccessDeniedException

AidenAtSea (approved):

src/main/java/software/amazon/cloudformation/exceptions/CfnUnauthorizedTaggingOperationException.java
ammokhov: We dont want to add custom message, this is supposed to behave more like an internal failure with static message

shrsrivz (approved):

aygold92 (approved):

src/main/java/software/amazon/cloudformation/exceptions/CfnUnauthorizedTaggingOperationException.java
aygold92: wait, why would we not want to have a normal message?

PR #429, opened 2023-10-30
feat: support Document type serialization

WinterYukky:
*Issue #, if available:*
`AWS::Kendra::DataSource` reverted support for `TemplateConfiguration`. This is because the template present in the `TemplateConfiguration` was a `Document` type that was not supported by the `cloudformation-cli-java-plugin`. It seems that `template` was implemented as a string type when it was initially implemented, so `callGraphs' type resolution failed. This investigation is the cause I personally discovered in the process of cloning and debugging the source and trying to register and deploy it in the CloudFormation Registry.
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-kendra/pull/107

*Description of changes:*
Add support Document type. 
I have confirmed that if `TemplateConfiguration.Template` is implemented by the Document type, it can at least be deployed with CloudFormation Registry.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -183,6 +183,10 @@
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
+            <artifactId>kendra</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
             <artifactId>apache-client</artifactId>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -185,7 +185,6 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>apache-client</artifactId>
         </dependency>
-        <!-- Test Dependencies -->
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>s3</artifactId>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -181,6 +181,10 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>cloudwatchlogs</artifactId>
         </dependency>
+        <dependency>
```
himanshs23, line 184: why we need to add this dependency?

himanshs23 (commented):
Other code looks good except for the dependency addition

pom.xml
WinterYukky: Because to test with actual SDK class like exinsting tests (e.g. `cloudwatchlogs.PutTargetsRequest`).
https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/429/files#diff-0df6d780ae71a8519e522b0d64d0792c1c10362f9b8bd9c0c3e983cf741ac6a9R62-R72

I should test using created class for test only instead new dependency?

WinterYukky: I change scope of dependency to test.

himanshs23 (approved):

SwaySway (approved):

